PR #4, opened 2015-08-26
Added new tech option; shortened prompt length to avoid line wrapping

andrewconnell:


robmhoward:
Nice 😀

jthake:
Boom. Thanks for adding that @andrewconnell , that's awesome!

andrewconnell:
You bet... good suggestion @robmhoward for the `manifest-only` option!

PR #31, opened 2015-09-09
added default values to manifest.xml files to avoid issues with uploading them to Add-in catalog [issue#29](/OfficeDev/generator-office/issues/29)

waldekmastykarz:


andrewconnell:
There are some issues that need to be addressed. 
- There's no comment in the PR that indicates what this is for... please update it. See [PR#8](/OfficeDev/generator-office/pull/8) for a reference how this is done.
- This PR deletes a lot of files that should be present... this must be fixed. For instance, the `.jscsrc` & `.jshintrc` files are missing from the root.
  - This might be due to the fact that the **dev** branch was out of sync. You should make your changes by branching off **master**. Please review the [contributing-workflow](/OfficeDev/generator-office/blob/master/docs/contributing-workflow.md) for the suggested workflow.
- If this is a code change, bump the version number accordingly in `package.json`. We use symver for versioning... if this adds functionality, bump the minor version. If it is a patch, bump the revision number.

Issue #29, opened 2015-09-09
Generator creates invalid manifest.xml file


waldekmastykarz:
The generator creates an invalid add-in manifest file. If you try to upload the manifest.xml directly after generating the add-in, the Apps for Office catalog in SharePoint Online will consider the manifest as invalid. In order to fix the issue you need to remove the empty `<SupportUrl/>` element from manifest.xml. After that you will be able to upload the manifest.xml file as expected.

andrewconnell:
When the `manifest.xml` file is built by the generator, it has some stuff in it that is designed for the user to act on... it's not perfect. For instance if you need regex matches in Outlook forms or to update the description and final URLs for the add-in's website.

One solution so it passes validation is to insert a dummy URL or insert the placeholder URL that the generator replaces. 

Thoughts?

waldekmastykarz:
How about doing something similar as for the `ProviderName` element or the `Description DefaultValue` attribute where the value is a placeholder that the dev has to replace?

andrewconnell:
Good suggestion... want to take it on?

waldekmastykarz:
Definitely!

PR #33, opened 2015-09-10
Added default values to manifest.xml files to avoid issues with uploading them to Add-in catalog

waldekmastykarz:
Added default values to manifest.xml files to avoid issues with uploading them to Add-in catalog as described in #29


andrewconnell:
Not sure why the travis build keeps failing... running the same commands locally works just fine & all tests pass. Have some house cleaning stuff to do on `package.json` so hopefully that will address it.

Issue #42, opened 2015-09-20
Add Angular + ADAL JS as technology


waldekmastykarz:
Add Angular + ADAL JS as technology to simplify building add-ins connected to Office 365. Add-in generated using this technology would contain all the manual steps described in the [ADAL JS manual](https://github.com/OfficeDev/generator-office/blob/master/docs/adal-config.md).

andrewconnell:
I like the approach provided this is a new tech option and not a change to the existing angular tech option. Want to take it on @waldekmastykarz >

jthake:
Let do it like how the angular generator has controller generators. check out script-base.js https://github.com/yeoman/generator-angular/blob/master/script-base.js
we can just have placeholder comments that file i/O can find and inject needed code.

waldekmastykarz:
Sure!

PR #69, opened 2015-10-13
Added support for ADAL technology #42

waldekmastykarz:
Added support for ADAL technology solving issue #42.


waldekmastykarz:
It seems like the coverage dropped due to the optional appId prompt which is available only for ng-adal and which is included in the `generators/app/index.js` file outside of the specific subgenerators for which we have tests.
@andrewconnell does that mean that we need to add tests for ng-adal in the `test/app.js` file to test the appId prompt or is the current situation acceptable?
What's surprising is that if I run `$ gulp test` locally the coverage shows 100% rather than the 97.685% reported by coveralls.

andrewconnell:
Don’t trust what is reported by the gulp task… it’s a known issue that it isn’t the best on reporting usage. Coveralls is the authority here. You can run it locally and see what’s being missed by the test and what could be added to improve the test coverage. You can do this locally or view the online report. For instance this shows we do not have a test for this line: https://coveralls.io/builds/3823801/source?filename=generators%2Fcontent%2Findex.js#L506

waldekmastykarz:
All right. Let me see if I can add some test to keep the coverage intact.

waldekmastykarz:
@andrewconnell would it be okay to keep the coverage intact by changing the tech of for example the Taskpane generator to `ng-adal` at https://github.com/OfficeDev/generator-office/blob/master/test/app.js#L73? This would keep the base test the same and at the same time would allow us to test the ng-adal path which we are not testing currently and which is causing the coverage drop.

waldekmastykarz:
With regard to the differences between the results of `$ gulp test` vs. `$ npm run-script test-travis`: it seems like we might be having the issue described in https://github.com/SBoudrias/gulp-istanbul/issues/37. Running `$ gulp test` shows a summary of 100% and 0 statements. If you run the travis test locally you get the same results as you get with coveralls on-line.

andrewconnell:
Sure… if the coverage doesn’t change then I don’t see what the problem would be with this

PR #125, opened 2016-01-15
Implements Outlook add-in commands

### LINKED ISSUE #55: Add support for add-in commands to mail add-in
We should show off the new add-in commands in the add-in that's generated.

http://blogs.msdn.com/b/exchangedev/archive/2015/08/31/introducing-add-in-commands.aspx
http://blogs.msdn.com/b/exchangedev/archive/2015/09/01/building-add-ins-with-commands.aspx

I have a demo add-in that shows the three command button types that's a straight HTML/CSS/JS project. It also uses Fabric: https://github.com/jasonjoh/command-demo

I'd propose using this in place of the demo add-in that's there now. We could prompt the user to include the command button types they are interested in and only include those types. I'm working on a prototype in my fork of this repo, I'll update once I have something to look at.

jasonjoh:
I implemented this as a subgenerator called ‘commands’. I designed it with two scenarios in mind:
- One of the existing subgenerators (‘mail’, ‘content’, etc) invokes it as part of their processing
- The user calls ‘commands’ directly to update an existing manifest

For full details see https://github.com/officedev/generator-office/issues/55#issuecomment-164003398

Resolves #55


jthake:
I just tested this. When you select HTML/JS option for Mail you do not get the add-in command questions like you do if you select Angular. is there a reason this hasn't been done? This change needs to be reflected on all the options for mail for consistency.

For manifest only - Is there a way where it prompts for function file, task pane url, icon url can have examples in brackets to make it more obvious what format it needs?

jasonjoh:
The HTML/JS sample doesn't prompt you because it supplies all of the values in an overrides.xml file. This is done so that the buttons that get created mirror the existing sample, creating similar functionality in a client that supports commands and one that doesn't.

The Angular sample doesn't have any sample functionality, so it doesn't use overrides.xml.

For manifest only, I can certainly add prompts for this. Is that something that needs to be done before we get it merged?

jthake:
I'd personally rather have a complete picture than have it work for just one scenario as it'll confuse users. How much work is it for this? I know this means its a bit more work. 

I think we need to drop the sample out of HTML so that this also can handle add-in commands. I will create an issue for that.

jasonjoh:
Well to be clear, the HTML option *does* handle add-in commands, it just doesn't ask the user for it. The subgenerator still generates all the entries in the XML for you. This was done to minimize how much we have to ask the user for and give the calling generator more control over the end result.

It might make sense to have a quick chat about this so there's no confusion :)

jthake:
Can we be consistent with the HTML questions and the Angular one and Manifest only one? Can we just assume that all ask the same questions?

jasonjoh:
Well, if we do #126, it would do that. The commands subgenerator doesn't choose to not prompt based on the tech, it chooses not to prompt based on the presence of a JSON object pre-defining all of the commands that gets passed by the mail generator. My concern about making it prompt for commands while the HTML sample is still there is that it creates a disjointed experience. OWA or Mac Outlook users would see the old UI with "Get Subject, Set Subject, etc", while Outlook 2016 users would see the generated buttons "UI-less Button 1", etc. 

Either way I think it leaves the HTML/JS scenario in an odd place until #126 renders it moot. We either have a sample with wildly different scenarios depending on client, or we have the potentially confusing lack of prompts. I can do either way, the amount of work is small.

PR #140, opened 2016-02-26
pull request to add default root path '.'

DevArvindsa:
Yo Office instructions does not work when you hit enter for the root path, we are missing default '.'


jthake:
Shipped in 0.6.4.

PR #143, opened 2016-02-29
changed gulp validate-xml behaviour for mail add-ins …

gavinbarron:
https verifciation now uses the correct object traversal to read the SourceLocation DefaultValue

This is only checks the first location of the form found. 

Note: I couldn't find any test that looked at the behavior of the tasks inside the gulpfiles output by the generator and frankly didn't know where to start


jthake:
Thanks @gavinbarron . I need to review a bunch of PRs and will get to this hopefully next week.

jthake:
Merged into Dev and under release 0.6.5 plans

PR #152, opened 2016-03-02
Remove extra commands from default generator flow

### LINKED ISSUE #10: Generator seems to hang waiting for user to hit Enter to finish
Using generator on Windows 10 in the following manner:

    yo office
    ? Project name (display name): yotest2
    ? Root folder of project? Default to current directory
     (C:\Users\jasonjoh\Source\Repos\mail-addins\yotest2), or specify relative path
      from current (src / public):
    ? Office project type: Mail Add-in (read & compose forms)
    ? Technology to use: HTML, CSS & JavaScript

After the generator does its work, and the final line is output:

    microsoft.office.js#2b215a298b bower_components\microsoft.office.js

The generator seems to hang, with the cursor just flashing. If you press Enter, it will exit.

Ideally the need to hit Enter can be removed, but if it can't, the tool should tell the user they need to hit Enter to exit.

pcostesi:
Rely on subgenerators/commands to get that info. We're keeping the
options in `app.js` so they transfer to subgenerators.

Fixes #10.

Tests: 
- created a new folder and ran "yo office" inside. Selected the default options, save for "task pane". Result: does not hang.
- created a new folder and ran "yo office:taskpane" inside. Ran with default options. Result: does not hang.
- created a new folder and ran "yo office:mail" inside. Ran with default options. Result: hangs.
- created a new folder and ran "yo office" inside. Ran with default options, save for "mail". Result: hangs.
- created a new folder and ran "to office:content". Ran with default options. Result: does not hang.

### DIFF
--- a/generators/app/index.js
+++ b/generators/app/index.js
@@ -41,13 +41,13 @@ module.exports = generators.Base.extend({
       desc: 'Office client product that can host the add-in',
       required: false
     });
-    
+
     this.option('extensionPoint', {
       type: String,
       desc: 'Supported extension points',
       required: false
     });
-    
+
     this.option('appId', {
       type: String,
       desc: 'Application ID as registered in Azure AD',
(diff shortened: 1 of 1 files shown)


jthake:
Merged into Dev and under release 0.6.5 plans

PR #179, opened 2016-09-14
Add SupportUrl element to Mail addin manifest file

### LINKED ISSUE #175: Mandatory 'SupportUrl' element is missing from manifest
Ideally the template would ask for the support URL and auto-populate such field.  Or at least generate a default value in the manifest.

vboctor:
This fix provides a default value similar to Description.  Maybe in the future, yo can ask for the support url and insert it as a value.

Fixes #175

### DIFF
--- a/generators/mail/templates/common/manifest.xml
+++ b/generators/mail/templates/common/manifest.xml
@@ -7,6 +7,7 @@
   <DefaultLocale>en-US</DefaultLocale>
   <DisplayName DefaultValue="<%= projectDisplayName %>" />
   <Description DefaultValue="[Outlook Add-in description]"/>
+  <SupportUrl DefaultValue="http://contoso.com/support" />
   <HighResolutionIconUrl DefaultValue="https://localhost:8443/images/hi-res-icon.png"/>
   <Hosts>
     <Host Name="Mailbox" />


generators/mail/templates/common/manifest.xml
```diff
@@ -7,6 +7,7 @@
   <DefaultLocale>en-US</DefaultLocale>
   <DisplayName DefaultValue="<%= projectDisplayName %>" />
   <Description DefaultValue="[Outlook Add-in description]"/>
+  <SupportUrl DefaultValue="http://contoso.com/support" />
```
beth-panx: <SupportUrl> element is only required if the add-in is built to be submitted to the Office Store. For internal use only, it will not be necessary. If you could add a comment to explain it and include a sample such as , I'd be happy to merge it to the source code.

beth-panx: SupportUrl element is only required if the add-in is built to be submitted to the Office Store. Some customers build their add-ins to be used internally only and this won't be necessary. If you could add a comment to it and provide an example such as `<!--<SupportUrl DefaultValue="[Outlook Add-in support url]">-->`, I'd be happy to merge it into the source code.

beth-panx:
SupportUrl element is only required if the add-in is built to be submitted to the Office Store. Some customers build their add-ins to be used internally only and this won't be necessary. If you could add a comment to it and provide an example such as `<!--<SupportUrl DefaultValue="[Outlook Add-in support url]">-->`, I'd be happy to merge it into the source code.

generators/mail/templates/common/manifest.xml
vboctor: The reason I added this entry is that the manifest validator was complaining on this field missing.  I can see other fields like Description where we have a "[Outlook Add-in description]" default value and Provider with "[Provider Name]".  So we can either have a place holder, e.g. "[Support URL]" or not fail the manifest validation when this field is not supplied or supplied but empty.  Thoughts?

P.S. Thanks for reviewing the PR.

beth-panx: No problem :) Thanks for using this tool.

I have been playing with Yo Office and found some issues with the validator itself. I am currently working on some of the changes so hopefully soon it won't complain about it. :p As of right now, the validation process is failing for me on some specific scenarios even after adding SupportUrl element to it. So at this point, I think either way could serve as a temporary workaround: a. add comment b. add SupportUrl.

beth-panx: Great point! I am putting this down as a task for the sprint. Will close this pr then :)

Issue #181, opened 2016-09-26
Outlook: Manifest from generator is missing IconUrl entry


jasonjoh:
Per Andrew this should be there.

beth-panx:
Resolved by pr #184

PR #184, opened 2016-10-05
Manifest element update

beth-panx:
Added IconUrl element to have a default value. Added comment for optional SupportUrl element. Updated tests to cover these changes.


beth-panx:
@jasonjoh would you like to review the changes I made that added IconUrl and optional SupportUrl elements to manifest files? This will close issue #175 and #181

PR #207, opened 2017-02-02
Updated installation instructions

Reezaali:


beth-panx:
Thanks Reeza! Looks good to me :)

PR #238, opened 2017-05-09
Adding support for generating projects exported from Script Lab

WrathOfZombies:
Generates projects for gists exported from Script Lab. Still in beta and needs work but the current iteration should work for most scenarios.


Zlatkovsky:
@WrathOfZombies , sorry, I somehow missed that this was ready to look at.

Let me review it in a few days.  In the meantime, could you:

1. Assign the merge to me
2. Add me as an admin on the project

Thanks!

akrantz:
This work would need to be re-done now given that we're changing generator-office to pull the projects from repositories.

PR #272, opened 2018-01-03
doc: Browsersync -> openssl

Hongbo-Miao:
I think we switched to openssl


MIchaelMainer:
We are still using BrowserSync to serve this up. We just aren't using the default BrowserSync certs since they no longer are for localhost.

Hongbo-Miao:
Oh okay

PR #276, opened 2018-01-09
add baseDir subproperty to bsconfig and update browser-sync version in package.json files

Rick-Kirkham:


casieber:
With #291 we have moved off of browser-sync fully, so this change is no longer needed. Closing.

Issue #269, opened 2017-12-18
Automate adding the dev cert authority


MIchaelMainer:
Currently, we must manually add the certificate to the cert store. This needs to happen once. This is an extra step that prevents the user from quickly getting started.

Here's an example implementation
https://github.com/Microsoft/web-build-tools/commit/407208e26283b9d87984f76a8cd07c6c7ff08447

LouMM:
Closing issue as per the latest fixes

PR #280, opened 2018-02-20
Fixed Typescript-React project generation failures

casieber:
-	Bumping all dependencies up to the latest versions
-	Reorganizing dependencies / devDependencies accordingly
-	Removing troublesome packages that were causing issues
-	Updating code to match standard patterns with the latest React versions
-	Pulled out BrowserSync since it’s capabilities are covered by the latest versions of webpack
-	Deleting redundant config JSON
-	Swapping SASS out for LESS (it was the node-sass module causing the python issues, as it has a hard requirement on node-gyp/python, which has [special required steps on Windows](https://github.com/nodejs/node-gyp#on-windows)).


casieber:
@akrantz 

>If webpack-dev-server has TLS certs, is it neccessary to generate/use specific certs?

No it is not neccessary to use custom certs, but since we used those certs in the past anyone who had installed them previously will already have them installed if they try to create a new office app with the new package. Switching to webpack-dev-server's generated certs would just mean some people might have to install new certs if they try to use the new office-generator.

I'm open to either course, just thought I would stick with what we had going before for now.

>If certs need to be generated, can the scripts trust the ca.cert if not already trusted?

The certs are pre-generated and committed as part of the repo (see src\app\templates\ts\base\certs\*), so there is no cert generation that happens on install.

akrantz:
@casieber There is still a user step to trust the generated ca.cert.  This can be a separate issue (https://github.com/OfficeDev/generator-office/issues/269) but it would be good to prompt and automatically install the ca.cert. 

If the webpack-dev-server default certs are already trusted, then perhaps it is simpler to not generate new certs.

The goal is to simplify the process by not having the user need to manually add the trusted cert.

casieber:
@akrantz There is also still a user step to trust webpack-dev-server default certs. So either route the user is still going to need to trust a cert. And even further, for webpack-dev-server, the certs are generated as part of server start, and not only that but they are deleted and regenerated on a cycle, meaning if we go with the default webpack-dev-server certs then the user will possibly have to trust certs more than once.

We could probably try to prompt for trusting our certs during template generation, but we would need to handle cross-platform scenarios which I don't have time to do right now so I'm going to defer that until later.

akrantz:
@casieber I noticed some miscellaneous lint warnings about spaces at the ends of lines and extra semicolons in App.tsx. You might want to go through the other files and double-check it. Otherwise, things look good to me.

casieber:
Looks like most of the lint errors existed before these changes, as you can see the lines mentioned in the above errors weren't touched, but I've gone ahead and fixed them as well.

jayrathi:
The resource.html file still has references to browser-sync. This needs to be updated

src/app/templates/ts/react/resource.html
```diff
@@ -9,7 +9,7 @@
...
+    <script type="text/javascript">  var appInsights=window.appInsights||function(config){    function i(config){t[config]=function(){var i=arguments;t.queue.push(function(){t[config].apply(t,i)})}}var t={config:config},u=document,e=window,o="script",s="AuthenticatedUserContext",h="start",c="stop",l="Track",a=l+"Event",v=l+"Page",y=u.createElement(o),r,f;y.src=config.url||"https://az416426.vo.msecnd.net/scripts/a/ai.0.js";u.getElementsByTagName(o)[0].parentNode.appendChild(y);try{t.cookie=u.cookie}catch(p){}for(t.queue=[],t.version="1.0",r=["Event","Exception","Metric","PageView","Trace","Dependency"];r.length;)i("track"+r.pop());return i("set"+s),i("clear"+s),i(h+a),i(c+a),i(h+v),i(c+v),i("flush"),config.disableExceptionTracking||(r="onerror",i("_"+r),f=e[r],e[r]=function(config,i,u,e,o){var s=f&&f(config,i,u,e,o);return s!==!0&&t["_"+r](config,i,u,e,o),s}),t    }({        instrumentationKey:"c36e2c25-f461-482c-9ead-bebc91e8b58b"    });           window.appInsights=appInsights;    appInsights.trackPageView("resource.html", "");</script>
 </head>
 
 <body class="<%= hostInternalName %>">
```
jayrathi: Need to update this file to remove references to browser-sync

casieber: Have updated

PR #299, opened 2018-04-17
Update generator prompts

TCourtneyOwen:
- Added new output param to specify folder if user wants it to be different than add-in name
- Removed prompt to specify folder name. Will default to add-in name for folder name.  If the user wishes to specify a different name they can use the output param
-Updated Typescript prompt so it now shows Typescript and JavaScript. This makes it clearer to the user what they are selecting if they don't select Typescript
- removed logic to show resource.html
- Changed logic so that we always call _postInstallHints at the end of yo office. ._postInstallHints mentions that a resource.html was created and provides other post-install info
- Added two new tests to new-project.ts
- Updated readme.md with new output param

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -59,7 +59,7 @@ The Microsoft Office client application that can host the add-in. The supported 
   - Optional
 
 #### `framework`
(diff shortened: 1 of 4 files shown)


TCourtneyOwen:
going to open new pull request with updated comments

PR #326, opened 2018-05-14
simplify yo office prompting; add test class that installs and build …

TCourtneyOwen:
- simplify yo office prompting
- change ordering of command line arguments (now projectType, name, host)
- add detailed help for yo office args (yo office --details)
- update existing tests to reflect the change in argument ordering
- add new test class that that installs and build projects for all hosts and project types


src/app/index.ts
```diff
@@ -92,164 +156,61 @@ module.exports = yo.extend({
...
-      let durationForManifestOnly = (endForManifestOnly - startForManifestOnly) / 1000;
-
-      /**
-       * Configure user input to have correct values
+       /**
+       * Configure project properties based on user input or answers to prompts
        */
       this.project = {
-        folder: answerForFolder.folder,
+        folder: this.options.name || answerForName.name || this.options.output,
```
akrantz: Shouldn't precendence be ```this.options.output```, then ```answerForName.name```, then ```this.options.name```?

src/app/index.ts
```diff
@@ -92,164 +156,61 @@ module.exports = yo.extend({
...
-      let durationForTs = (endForTs - startForTs) / 1000;
-      if (!(this.options.js == null)) {
-        this.project.ts = !this.options.js;
-      }
-      else {
-        this.project.ts = answerForTs.ts || false;
-      }
-      if (this.options.framework === 'react') {
-        this.project.ts = true;
+      if (this.options.js){
```
akrantz: nit: space between ```)``` and ```{```

I see this other places as well...

src/app/index.ts
```diff
@@ -260,14 +221,20 @@ module.exports = yo.extend({
       this.project.projectInternalName = _.kebabCase(this.project.name);
       this.project.projectDisplayName = this.project.name;
       this.project.projectId = uuid();
-      this.project.hostInternalName = _.toLower(this.project.host);
-
-      if (this.project.folder) {
-        this.destinationRoot(this.project.projectInternalName);
+      if (this.project.projectType !== excelFunctions){
```
akrantz: nit: reverse if so it uses === and swap if and else clauses.

src/app/index.ts
```diff
@@ -260,14 +221,20 @@ module.exports = yo.extend({
...
+        this.project.hostInternalName = this.project.host;
       }
+      else {
+        this.project.hostInternalName = excelFunctions;
+      }      
+      this.destinationRoot(this.project.folder);
+
+      /** Check to to see if destination folder already exists. If so, we will exit and prompt the user to provide
+      a different project name or output folder */
+      this._projectFolderExists();
```
akrantz: This should really be whether the folder is not empty. If not empty, prompt the user for another location. If empty or if does not exist, then it is OK to create/use it.

src/app/index.ts
```diff
@@ -260,14 +221,20 @@ module.exports = yo.extend({
...
 
       let duration = this.project.duration;
-      insight.trackEvent('App_Data', { AppID: this.project.projectId, Host: this.project.host, Framework: this.project.framework, isTypeScript: this.project.ts.toString() }, { duration });
+      insight.trackEvent('App_Data', { AppID: this.project.projectId, Host: this.project.host, ProjectType: this.project.projectType/* , isTypeScript: this.project.scriptType = 'Typescript' */ }, { duration });
```
akrantz: Fix = to be === and use typescript constant.  ``` this.project.scriptType === typescript ```

src/app/index.ts
```diff
@@ -276,42 +243,54 @@ module.exports = yo.extend({
   writing: {
     copyFiles: function () {
       try {
-        let language = this.project.ts ? 'ts' : 'js';
+        let language = this.project.scriptType  === typescript ? 'ts' : 'js';
```
akrantz: You might want to change this.project.scriptType to this.project.language. For now, we only have typescript/javascript, but it would be more extensible to other languages later. It's not strictly necessary.

src/app/index.ts
```diff
@@ -276,42 +243,54 @@ module.exports = yo.extend({
...
+        if (!this.project.isManifestOnly && !this.project.isExcelFunctionsProject) {
           this.log('\n----------------------------------------------------------------------------------\n');
-          this.log(`      Creating ${chalk.bold.green(this.project.projectDisplayName)} add-in using ${chalk.bold.magenta(language)} and ${chalk.bold.cyan(this.project.framework)}\n`);
+          this.log(`      Creating ${chalk.bold.green(this.project.projectDisplayName)} add-in at ${chalk.bold.magenta(this._destinationRoot)} for ${chalk.bold.yellow(this.project.host)} using ${chalk.bold.magenta(language)}\n`);
           this.log('----------------------------------------------------------------------------------\n\n');
         }
-        else {
-          this.log('----------------------------------------------------------------------------------\n');
-          this.log(`      Creating manifest for ${chalk.bold.green(this.project.projectDisplayName)} add-in\n`);
+        else if (this.project.isExcelFunctionsProject) {
```
akrantz: Do we want/need a different banner here for an Excel custom functions project?

src/app/index.ts
```diff
@@ -276,42 +243,54 @@ module.exports = yo.extend({
   writing: {
     copyFiles: function () {
       try {
-        let language = this.project.ts ? 'ts' : 'js';
+        let language = this.project.scriptType  === typescript ? 'ts' : 'js';
 
         /** Show type of project creating in progress */
-        if (this.project.framework !== 'manifest-only') {
+        if (!this.project.isManifestOnly && !this.project.isExcelFunctionsProject) {
```
akrantz: The logic would be better reversed:

if (isManifestOnly) {
   …
} else if (isExcelCustomFunctionsProject) {
  ….
} else {
  …
}

This is why it is better to make if statements use positive logic, so you avoid double-negatives.  :-)

src/app/index.ts
```diff
@@ -369,6 +361,56 @@ module.exports = yo.extend({
...
+        else {
+          this.options.projectType = element;
+        }
+        return true;
+      }
+    }
+    return false;
+  },
+
+_projectFolderExists: function ()
```
akrantz: The function is checked if the project folder is not empty (which is good). However, you should update the function name to match what it really does.

src/app/index.ts
```diff
@@ -92,164 +156,61 @@ module.exports = yo.extend({
       }];
       let answerForName = await this.prompt(askForName);
       let endForName = (new Date()).getTime();
-      let durationForName = (endForName - startForName) / 1000;
+      let durationForName = (endForName - startForName) / 1000; 
 
-      /** office client application that can host the addin */
+      /** askForHost will be triggered if no project name was specified via the command line Host argument, and the Host argument
+       * input was in fact valid, and the project type is not ExcelexcelFunctions */
```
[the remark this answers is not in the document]
TCourtneyOwen: fixed

src/app/index.ts
TCourtneyOwen: The reason I created a separate banner for custom functions is that custom functions aren't strictly associated with a particular host type (i.e users are not required or prompted for a host type if projectType == excel-functions

src/app/index.ts
TCourtneyOwen: changing this will require pretty extensive changes throughout index.ts and the test classes. Not sure if it's worth it for now

Issue #330, opened 2018-05-29
React projects don't appear to be building correctly, and thus we can't properly debug them


TCourtneyOwen:
While working on a fix for VSO.2088313 (Yo Office  - Typescript source mapping does not work when debugging). I noticed that we run 'npm run build' for React projects,  we end up generating very wonky names for the js and source-map files.  I think this may contribute to the fact that when I try to debug React project in VS2017, I am unable to see the .ts file at all in Visual Studio.  I suspect this may be related to the specialized webpack config files that we copy over for React, which differ from those we use for Angular and Jquery typescript projects.

TCourtneyOwen:
This is no longer an issue - closing as resolved

PR #331, opened 2018-05-29
-Update webpack.config.js so that the source map files are built when…

TCourtneyOwen:
… running 'npm run build' for typescript projects

- Exercised scenario cited in https://office.visualstudio.com/DefaultCollection/OC/Excel/_workitems/edit/2088313 to ensure we now work as expected and that we can now set a breakpoint in index.ts and that we hit that breakpoint when running code for the add-in
- I was able to confirm this fix for Jquery nd Angular typescript projects, but it seems like we have a general problem with React projects seemingly related to the specialized webpack config files that we copy over for React.  When we run 'npm run build' for React projects, we end up generating very wonky names for the js and source-map files.  I think this may contribute to the fact that when I try to debug React project in VS2017, I am unable to see the .ts file at all. I opened https://github.com/OfficeDev/generator-office/issues/330 to track this issue with React typescript projects

### DIFF
--- a/src/app/templates/ts/base/webpack.config.js
+++ b/src/app/templates/ts/base/webpack.config.js
@@ -1,6 +1,7 @@
 const HtmlWebpackPlugin = require('html-webpack-plugin');
 
 module.exports = {
+    devtool: 'source-map',
     entry: {
         app: './src/index.ts',
         'function-file': './function-file/function-file.ts'


akrantz:
@casieber can help with the react webpack config.

TCourtneyOwen:
@akrantz Are you OK with me pushing this up to the main repo? I still wasn't able to get end-to-end debugging working for React projects (even though we are in fact correctly creating the source-map file for React), but I don't think that should stop us from checking in this change.  The React debug issue is tracked by https://github.com/OfficeDev/generator-office/issues/330

PR #334, opened 2018-06-26
-Fix for Bug 2370591 so that we now provide better project names in Y…

TCourtneyOwen:
…o Office UI prompt

- Removed some stale code from index.ts that wasn't being used anymore

### DIFF
--- a/src/app/config/projectProperties.json
+++ b/src/app/config/projectProperties.json
@@ -7,20 +7,14 @@
                     "repository": ""},
                 "javascript": {
                     "repository": ""}
-                },
-            "typescript": true,
-            "javascript": true
+                }        
         },
         "excel-functions": {
             "displayname": "Excel Custom Functions Add-in project (Preview: Requires the Insider channel for Excel)",
             "templates": {
-                "typescript": {
-                    "repository": ""},
                 "javascript": {
                     "repository": ""}
-                },
-            "typescript": false,
-            "javascript": true
+                }
         },
         "jquery": {
             "displayname": "Office Add-in project using Jquery framework",
@@ -29,31 +23,21 @@
                     "repository": ""},
                 "javascript": {
                     "repository": ""}
-                },
-            "typescript": true,
-            "javascript": true
+                }
         },
         "manifest": {
             "displayname": "Office Add-in containing the manifest only",
             "templates": {
-                "typescript": {
-                    "repository": ""},
-                "javascript": {
+                "manifest-only": {
                     "repository": ""}
-                },
-            "typescript": false,
-            "javascript": false
+                }
         },
         "react": {
             "displayname": "Office Add-in project using React framework",
             "templates": {
                 "typescript": {
-                    "repository": ""},
-                "javascript": {
                     "repository": ""}
-                },
-            "typescript": true,
-             "javascript": false
+                }
             }
         },
         "hostTypes": {
@@ -78,5 +62,6 @@
         "word": {
             "displayname": "Word",
             "repository": ""}
-    }
-}
+        }
+ }
+
--- a/src/app/config/projectProperties.json
+++ b/src/app/config/projectProperties.json
@@ -1,0 +1,68 @@
+{
+    "projectTypes": {
+        "angular": {
+            "name": "angular",
+            "displayname": "Office Add-in project using Angular framework",
+            "repolocation": "",
+            "typescript": true,
+            "javascript": true
+        },
+        "excel-functions": {
+            "name": "excel-functions",
+            "displayname": "Excel Custom Functions Add-in project (Preview: Requires the Insider channel for Excel)",
+            "repolocation": "",
+            "typescript": true,
+            "javascript": true
+        },
+        "jquery": {
+            "name": "jquery",
+            "displayname": "Office Add-in project using Jquery framework",
+            "repolocation": "",
+            "typescript": true,
+            "javascript": true
+        },
+        "manifest": {
+            "name": "manifest",
+            "displayname": "Office Add-in containing the manifest only",
+            "repolocation": "",
+            "typescript": true,
+            "javascript": true
+        },
+        "react": {
+            "name": "react",
+            "displayname": "Office Add-in project using React framework",
+            "repolocation": "",
+            "typescript": true,
+             "javascript": false
+            }
+        },
+    "hostTypes": {
+        "excel": {
+            "name": "excel",
+            "displayname": "Excel",
+            "repolocation": ""
+        },
+        "onenote": {
+            "name": "onenote",
+            "displayname": "OneNote",
+            "repolocation": ""
+        },
+        "outlook": {
+            "name": "outlook",
+            "displayname": "Outlook",
+            "repolocation": ""},
+        "powerpoint": {
+            "name": "powerpoint",
+            "displayname": "PowerPoint",
+            "repolocation": ""
+        },
+        "project": {
+            "name": "project",
+            "displayname": "Project",
+            "repolocation": ""},
+        "word": {
+            "name": "word",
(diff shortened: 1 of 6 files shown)


src/app/index.ts
```diff
@@ -433,16 +430,21 @@ function getDirectories(root) {
...
   }
+}
+
+function updateProjectNames(projectTemplate)
+{
+  if (projectTemplate == 'Manifest'){
+    projectTemplate = 'Office Add-in ' + projectTemplate + ' only';
+  }
+  else{
+    projectTemplate = 'Office Add-in using ' + projectTemplate + ' framework';
```
akrantz: Custom Functions is not a framework. Seems better to have a function that maps each projectTemplate string to a project display name string. Instead of updateProjectNames(projectTemplate) I would propose getProjectDisplayName(projectTemplate) and use a switch that just returns each string. Don't worry about trying to use projectTemple as part of the string formatting.  Even better would be to have this be data driven from a json file so adding a new one wouldn't require changing the function but just returning the mapping.

src/test/build-all-projects.ts
```diff
@@ -27,49 +29,55 @@ describe('Setup test environment for Yo Office build tests', () => {
...
+        if (parsedProjectJsonData.projectTypes[projectTemplates[j]].javascript)
+        {
+            describe('Install and build projects', () => {
+                let projectType = projectTemplates[j];        
+                let host = hostsTemplates[i];
+                let scriptType = javascript;
+                let projectName = projectType + host + javascript;
+                let projectFolder = path.join(__dirname, '/', projectName);          
+            
+                describe(stringBuildStart +  host + space + projectType + space + typescript, () => {
```
MsCesarBerard: >typescript [](start = 82, length = 10)

Shouldn't this be **javascript**?

src/test/build-all-projects.ts
```diff
@@ -27,49 +29,55 @@ describe('Setup test environment for Yo Office build tests', () => {
...
+                        _buildProject(projectFolder, projectType);
+                        done();                    
+                      });
+                  }); 
+                });
+            }
         }
-    }    
-    
+    }
```
MsCesarBerard: >} [](start = 4, length = 1)

nit: I think that the indentation is not correct (this happens as well in the **for** above)

src/test/build-all-projects.ts
```diff
@@ -27,49 +29,55 @@ describe('Setup test environment for Yo Office build tests', () => {
 // Build Typescript project types for all supported hosts
 for (var i = 0; i < hostsTemplates.length; i++)
 {
-    for (var j = 0; j < tsTemplates.length; j++)
+    for (var j = 0; j < projectTemplates.length; j++)
```
MsCesarBerard: >for (var j = 0; j < projectTemplates.length; j++) [](start = 4, length = 49)

I think that the build Typescript and Javascript logic can be refactored and execute it within this loop (instead of creating two more nested loops), the only difference seems to be the use of the **javascript** and **typescript** const, maybe you can make a function which receives the args and then make the rest of the logic using it, e.g. 

if (parsedProjectJsonData.projectTypes[projectTemplates[j]].typescript){
    build(projectTemplates[j], hostTemplates[i], typescript);
}
if (parsedProjectJsonData.projectTypes[projectTemplates[j]].javascript){
    build(projectTemplates[j], hostTemplates[i], javascript);
}

src/app/config/projectProperties.json
```diff
@@ -0,0 +1,68 @@
+{
+    "projectTypes": {
+        "angular": {
+            "name": "angular",
```
akrantz: Each property within "projectTypes" is the name of the project type, so it is redundant to have it also have the same thing as the "name" property within the object.

src/app/config/projectProperties.json
```diff
@@ -0,0 +1,68 @@
...
+            "displayname": "Office Add-in project using Angular framework",
+            "repolocation": "",
```
akrantz: "repository" would be simpler than "repolocation".

akrantz: You'll really need a repository location per ts/js language. I think it would be better to have a languages property with a sub-object with "javascript" and "typescript" which for each has "repository".

akrantz: Something like this:
```
propertyTypes: {
     "excel-functions": {
          "displayName": "Excel Custom Functions Add-in",
          "templates: {
             "typescript": {
                    "repository: "https://www.github.com/OfficeDev/office-addin-templates-excel-functions"
             },
             "javascript": {
                    "repository: "https://www.github.com/OfficeDev/office-addin-templates-excel-functions"
             }
    }
```

src/app/config/projectProperties.json
TCourtneyOwen: Fixed locally

src/test/build-all-projects.ts
TCourtneyOwen: Fixed locally

src/app/config/projectProperties.json
```diff
@@ -1,68 +1,82 @@
...
             "typescript": true,
             "javascript": true
         },
         "excel-functions": {
-            "name": "excel-functions",
             "displayname": "Excel Custom Functions Add-in project (Preview: Requires the Insider channel for Excel)",
-            "repolocation": "",
-            "typescript": true,
+            "templates": {
+                "typescript": {
```
akrantz: You could simplify this by using whether a "typescript" or "javascript" template is available, and not need the bool property. This would mean to have only javascript template right now for excel-functions project type, and not have either for manifest project type. The others would have both.

src/app/config/projectsJsonData.ts
```diff
@@ -28,19 +25,16 @@ export default class projectsJsonData{
...
               return true;
             }
           }
           return false;
         }
     }
 
-    getProjectDisplayNames(projectTemplate){
-      return this.m_projectJsonData.projectTypes[_.toLower(projectTemplate)].displayname;
+    getProjectDisplayNames(projectType: string){
```
akrantz: This only returns one name, so the function name should be singular.

src/app/config/projectsJsonData.ts
```diff
@@ -72,4 +66,15 @@ export default class projectsJsonData{
       }
       return hosts;
     }
+
+    normalizeHostNameFromInput(input: string)
+    {
+      for (let key in this.m_projectJsonData.hostTypes)
```
akrantz: This function take "input" parameter which is matched against the known host key names, and the displayName is returned for the matching host. From that you could improve the naming of the function/parameter. For example: getHostDisplayName(hostKey: string).

src/test/build-all-projects.ts
```diff
@@ -13,78 +16,92 @@ const ts = '--ts';
...
-            let host = hostsTemplates[i];
-            let scriptType = typescript;
-            let projectName = projectType + host + typescript;
-            let projectFolder = path.join(__dirname, '/', projectName);          
-        
-            describe(stringBuildStart +  host + space + projectType + space + typescript, () => {
-                it(stringBuildSucceeds,function(done){  
-                    _generateProject(projectType, projectName, host, projectFolder, scriptType);
-                    _buildProject(projectFolder, projectType);
+        let host = hostsTemplates[i].toLowerCase();
```
MsCesarBerard: >let host = hostsTemplates[i].toLowerCase(); [](start = 8, length = 43)

nit: This can be executed in the **for** above, so this operation is not executed during the iterations of the project templates.

TCourtneyOwen: fixed locally

TCourtneyOwen:
Tests now don't timeout after removing testing against stale Node-js build 6.9.1

PR #338, opened 2018-07-09
Remove step to delete project folders from build-all-projects.ts to s…

TCourtneyOwen:
…ee if that helps the test run faster so we don't hit timeouts in TravisCI

### DIFF
--- a/src/test/build-all-projects.ts
+++ b/src/test/build-all-projects.ts
@@ -95,7 +95,7 @@ function _buildProject(projectFolder, projectType)
         shell.cd(__dirname);
         
         // do clean-up after test runs
-        _deleteFolderRecursively(projectFolder);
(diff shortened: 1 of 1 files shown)


akrantz:
@TCourtneyOwen Does it make a difference? Will the files still be cleaned up later?

TCourtneyOwen:
No this didn't fix the issue. The install/build tests are timing out where we are running tests against Node-Js 6.9.1, but do not timeout on either Linus or Mac when running against the latest stable build (10.6.0).  Frankly I don't know when we have our tests running against such an old version of Node-Js in addition to the latest stable build, but that's what we currently have slated in .travis.yml.  Perhaps we can just remove the 6.9.1 config from .travis.yml. I am pretty sure that will address the problem

PR #343, opened 2018-07-30
Proof-of-concept to show how we can clone project files from a separa…

TCourtneyOwen:
…te repro for each project type.

NOTE: This change id dependent on https://github.com/OfficeDev/Excel-Custom-Functions/pull/24.  We won't be able to make this change until we have updated https://github.com/OfficeDev/Excel-Custom-Function  to be consistent with the files we install as part of Yo Office custom functions.
- Clone https://github.com/OfficeDev/Excel-Custom-Function instead of copy files from src/app/templates/excel-custom-functions-preview
- Completly remove src/app/templates/excel-custom-functions-preview


src/app/index.ts
```diff
@@ -259,10 +260,11 @@ module.exports = yo.extend({
         
         if (this.project.isExcelFunctionsProject)
         {
-          this.fs.copyTpl(this.templatePath(`excel-custom-functions-preview/**`), this.destinationPath(), templateFills, null, { globOptions: { ignore: `**/*.placeholder` }});
+          // copy over custom functions files from Excel Custom Functions repo (this is the model we would like to have for all projects).
+          git.Clone("https://github.com/OfficeDev/Excel-Custom-Functions", this.destinationPath());
```
akrantz: This would ultimately make a .git directory in the destination. It would be good to know if we could make an option to clone but without creating the .git directory (thus just being a copy of the files themselves).

It could be considered a feature to have this as a git clone because you could potentially merge in fixes from origin later. But I could also see where someone might not want the .git stuff. I suppose they could just delete the .git folder afterwards if it isn't desired.

src/app/index.ts
```diff
@@ -259,10 +260,11 @@ module.exports = yo.extend({
         
         if (this.project.isExcelFunctionsProject)
         {
-          this.fs.copyTpl(this.templatePath(`excel-custom-functions-preview/**`), this.destinationPath(), templateFills, null, { globOptions: { ignore: `**/*.placeholder` }});
+          // copy over custom functions files from Excel Custom Functions repo (this is the model we would like to have for all projects).
+          git.Clone("https://github.com/OfficeDev/Excel-Custom-Functions", this.destinationPath());
+
+          // we can ultimatey create a repo with basefiles as well such as certs
           this.fs.copy(this.templatePath(`${language}/base/certs`), this.destinationPath('certs'), { globOptions: { ignore: `**/*.placeholder` }});
```
akrantz: Wouldn't we need the base stuff to be part of the git project? If so they wouldn't be copied here.

src/app/config/projectsJsonData.ts
```diff
@@ -77,4 +77,20 @@ export default class projectsJsonData{
       }
       return undefined;
     }
+
+    getProjectTemplateRepository(projectTypeKey: string, scriptType: string)
+    {
+      for (let key in this.m_projectJsonData.projectTypes)
+      {
+        if (_.toLower(projectTypeKey) == key){
+          if (scriptType == 'javascript'){
```
akrantz: Why not just use scriptType as a key in:

`return this.m_projectJsonData.projectTypes[key].templates[scriptType].repository; `

src/app/index.ts
```diff
@@ -74,7 +74,7 @@ module.exports = yo.extend({
     }
     let message = `Welcome to the ${chalk.bold.green('Office Add-in')} generator, by ${chalk.bold.green('@OfficeDev')}! Let\'s create a project together!`;
     this.log(yosay(message));
-    this.project = {};
+    this.project = {};
```
akrantz: nit: remove extra spaces added

src/app/index.ts
```diff
@@ -254,15 +255,16 @@ module.exports = yo.extend({
         let language = this.project.scriptType === typescript && !this.project.isExcelFunctionsProject  ? 'ts' : 'js';
         const starterCode = generateStarterCode(this.project.host);
         const templateFills = Object.assign({}, this.project, starterCode);
+        let jsonData = new projectsJsonData(this.templatePath()); 
 
         this._projectCreationMessage();
         
         if (this.project.isExcelFunctionsProject)
```
akrantz: While this is OK for now, you could also structure the code to see if there is a repository for the project type. If not (because null is returned), then it could fall back to the existing template logic in the else clause. That would make it for the next project, we just need to update the projects json with a repository path to use it and not rely on a special case in code for custom functions project.

akrantz:
I do not think the manifest is customized based on yo office inputs, such as project name. I think we may need to make an office-addins-customize-manifest script which can apply changes to an existing manifest using command line input options for generating a new add-in guid, specifying the name, etc.

src/app/config/projectsJsonData.ts
TCourtneyOwen: Duh. Good call :-)

src/app/index.ts
TCourtneyOwen: I think it might be easier to have a separate base repo so we only have to update the base files (e,g, certs, asset files)  one time rather than having to do it numerous times in each and every project repo

src/app/index.ts
TCourtneyOwen: I will add logic to fall back to existing template if getProjectTemplateRepository returns undefined

PR #351, opened 2018-09-07
I created two branches off of the Excel Custom Functions repo:

TCourtneyOwen:
Preview May 2018  this contains the current custom functions template and matches whats in master
Preview_Refresh_Sept_2018  this contains the custom functions with the updated CDN, as well as the replacement of OfficeExtension.Promise with just Promise for the async function example

I have made the updates in Yo Office so that we now show the following options:
  Office Add-in project using Angular framework
  Excel Custom Functions Add-in project (May 2018 Preview: Requires the Insider channel for Excel)
  Excel Custom Functions Add-in project (September 2018 Preview Refresh: Requires the Insider channel for Excel)
  Office Add-in project using Jquery framework
  Office Add-in containing the manifest only
  Office Add-in project using React framework

We will pull the template from the correct custom functions branch depending on which option the user chooses.  I will be sending out that PR is  a few.

We can make any additional changes to the Preview_Refresh_Sept_2018 branch independent from Yo Office, which is of course ideal and a major reason why we made this first initial move to having separate repos for each project template.

### DIFF
--- a/src/app/config/projectProperties.json
+++ b/src/app/config/projectProperties.json
@@ -4,54 +4,50 @@
             "displayname": "Office Add-in project using Angular framework",
             "templates": {
                 "typescript": {
-                     "repository": "",
-                    "branch": "master"},
+                    "repository": ""},
                 "javascript": {
-                    "repository": "",
-                    "branch": "master"}
-                }       
-        },
-        "excel-functions": {
-            "displayname": "Excel Custom Functions Add-in project (May 2018 Preview: Requires the Insider channel for Excel)",
-            "templates": {
-                "javascript": {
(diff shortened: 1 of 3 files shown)


src/app/config/projectProperties.json
```diff
@@ -4,39 +4,54 @@
             "displayname": "Office Add-in project using Angular framework",
             "templates": {
                 "typescript": {
-                    "repository": ""},
+                     "repository": "",
+                    "branch": "master"},
```
akrantz: Instead of having to include "branch": "master" for everything, it could use "master" as the default if not specified, and only require "branch" when specifying something else. The code should be able to handle "branch" not being listed in the json object.

Issue #354, opened 2018-09-17
generator cuts out of process at "What do you want to name your add-in"


ivan7707:
On my windows 10 new install, I was having an issue where I would start the generator, choose the project type, then the generator would automatically skip the "What do you want to name your add-in" line, and end the generator session.

This happens when I choose the options 2,4 and 5 on the list of project types.  For the Angular and Jquery options, the generator allows me to input a name for the add-in.

Even if I use the command line argument yo office react, it shows the name your add-in line, but I cannot type anything in, and it cuts out of the process itself after about 5 seconds. 

Windows 10.0.17134.285
Node: 10.10.0
Python 3.7.0

TCourtneyOwen:
Thanks for reporting this issue. We think this may be an issue with your version of Node, as a couple others have reported this as well with Node 10.10.0. We will look into this problem and try and get a fix out ASAP

-Courtney

ivan7707:
Thanks Courtney.  

Let me know if I can help out with logs, or testing of any type.   

Would you like me to downgrade to a previous version of Node to test?

Ivan

TCourtneyOwen:
Hi Ivan,

I should have a fix out this afternoon. I was also able to repro with node 10.10.0 on one of my machines. However I wasn't able to repro on another machine with the same version of Node.  

I will update this issue once I publish the new version

Thanks,

Courtney

ivan7707:
Thank you Courtney.  I am really looking forward to building some fun stuff with this module.

ivan7707:
The merge worked.  Thank you Courtney.

PR #356, opened 2018-09-17
Temportary fix for issue causing Yo Office to end prematurely on Node…

TCourtneyOwen:
… v10.10.0 (see https://github.com/OfficeDev/generator-office/issues/354)

- Still not sure why we bail out of Yo Office on Node 10.10.0 and don't on earlier versions of Node after we do the check for the script types available for the specified project type. For now, will just bypass the prompt for script type if promptForScriptType is false
- Also updated .travis.yml so we test on stable Node build as well as an earlier build

I think we should get this fix out ASAP because more people are starting to run into this

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ cache:
   - node_modules
 node_js:
   - 'stable'
+  - '8.12.0'
 before_script: "npm run build"
 script: "npm run test"
 os:
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,6 @@ cache:
   - node_modules
 node_js:
   - 'stable'
-  - '8.12.0'
 before_script: "npm run build"
 script: "npm run test"
 os:
--- a/src/app/index.ts
+++ b/src/app/index.ts
@@ -118,8 +118,8 @@ module.exports = yo.extend({
 
       // Determine if we should prompt for script type. This is to address a bug that is causing Yo Office to crash in Node v10.10.0 - Issue #354
       // This is a temportary fix - we need to clean up the code in the next major version of the generator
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -8,6 +8,7 @@ cache:
   - node_modules
 node_js:
   - 'stable'
+  - '8.12.0'
```
TCourtneyOwen: this node version doesn't appear to be available for Unix so tests are failing when try to run against this version in TravisCI

src/app/index.ts
```diff
@@ -116,21 +116,24 @@ module.exports = yo.extend({
       || (this.options.projectType != null && _.toLower(this.options.projectType) == excelCustomFunctions)) { 
         isExcelFunctionsProject = true; }
 
-      /* askForTs and askForProjectType will only be triggered if the js param is null, it's not a Manifest project,
-       * it's not an ExcelexcelFunctions project and the project type exists for both script types */      
+      // Determine if we should prompt for script type. This is to address a bug that is causing Yo Office to crash in Node v10.10.0 - Issue #354
+      // This is a temportary fix - we need to clean up the code in the next major version of the generator
+      let promptForScriptType = !isManifestProject && this.options.js == null  && this.options.ts == null && (this.options.projectType != null && jsonData.projectBothScriptTypes(this.options.projectType)
```
akrantz: Use === and !== for strict type checking.

src/app/index.ts
```diff
@@ -213,7 +216,7 @@ module.exports = yo.extend({
         projectType: _.toLower(this.options.projectType) || _.toLower(answerForProjectType.projectType),
         isManifestOnly: isManifestProject,
         isExcelFunctionsProject: isExcelFunctionsProject,
-        scriptType: answerForScriptType.scriptType
+        scriptType: answerForScriptType != undefined ? answerForScriptType.scriptType : undefined
```
akrantz: use !== and put parenthesis around clause before ? for clarity.

PR #364, opened 2018-09-28
fix(generate): write files in async task - fixes #363

### LINKED ISSUE #363: creating excel-functions fails because async method is treated as sync
The following error happens with `generator-office@1.1.17` 

internally `git.clone` is used which is async, but the `writing` method calls `_copyProjectFiles` sync. The following step in workflow *installing dependencies* throws because clone operation isn't finished yet.

This affects #360 and #362

ThorstenHans:
When templates come from an external repo, git.clone is used to pull the template. This method is async, so the generator has to wait until cloning actually finished

### DIFF
--- a/src/app/index.ts
+++ b/src/app/index.ts
@@ -181,7 +181,15 @@ module.exports = yo.extend({
   },
 
   writing: function () {
-    this._copyProjectFiles();
+    const done = this.async();
+    this._copyProjectFiles()
+    .then(() => {
+      done();
+    })
+    .catch((err)=>{
+      insight.trackException(new Error('Installation Error: ' + err));
+      process.exitCode = 1;
+    });
   },
 
   install: function () {
@@ -253,6 +261,7 @@ module.exports = yo.extend({
 
   _copyProjectFiles()
   {
+    return new Promise((resolve, reject) => {
       try {
         let language = this.project.scriptType === typescript ? 'ts' : 'js';
         const starterCode = generateStarterCode(this.project.host);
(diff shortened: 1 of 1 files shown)


TCourtneyOwen:
While I personally haven't hit any issues related to the cloning process, this change looks safe and an improvement to what we currently have. I know @akrantz was thinking that we could perhaps just have the cloning portion wrapped up in a promise, but I am personally OK with @ThorstenHans approach

src/app/index.ts
```diff
@@ -181,7 +181,15 @@ module.exports = yo.extend({
   },
 
   writing: function () {
-    this._copyProjectFiles();
+    const done = this.async();
+    this._copyProjectFiles()
+    .then(() => {
```
akrantz: This can be simplified as `.then(done)`

src/app/index.ts
```diff
@@ -181,7 +181,15 @@ module.exports = yo.extend({
   },
 
   writing: function () {
-    this._copyProjectFiles();
+    const done = this.async();
+    this._copyProjectFiles()
+    .then(() => {
+      done();
+    })
+    .catch((err)=>{
```
akrantz: nit: add spaces around `=>`

src/app/index.ts
```diff
@@ -265,7 +274,13 @@ module.exports = yo.extend({
         // Copy project template files from project repository (currently only custom functions has its own separate repo)
         if (projectRepoBranchInfo.repo)
         {
-          git().clone(projectRepoBranchInfo.repo, this.destinationPath(), ['--branch', (projectRepoBranchInfo.branch) ? projectRepoBranchInfo.branch : 'master']);
+          git().clone(projectRepoBranchInfo.repo, this.destinationPath(), ['--branch', (projectRepoBranchInfo.branch) ? projectRepoBranchInfo.branch : 'master'], (err) => {
+            const gitFolder = this.destinationPath() + '/.git';
+            if (fs.existsSync(gitFolder)){
+              helperMethods.deleteFolderRecursively(gitFolder);
+            }
+            return err? reject(err): resolve();
```
akrantz: nit: uses spaces on both sides of `?` and `:`.
 `return err ? reject(err) : resolve();`

TCourtneyOwen: Adam, I will make these changes since I have incorporated these changes into another change I am about to send out that calls into office-addin-manifest to modify the manifest guid and display name

akrantz: @TCourtneyOwen I think it would be better to merge this first, and then we can merge it into your manifest change branch.

TCourtneyOwen: OK, that's fine, but I am going to make these changes as opposed to waiting on Thorsten

PR #367, opened 2018-10-06
add fabric tags

schjan:
Without the <Fabric> Tag it seems that fonts and some CSS styles aren't loaded properly


schjan:
Thx for approving but I think this PR doesn't make sense anymore. The changed file is deleted on master and I didn't use this project this year so I don't know if it is still relevant.

So closing it now

lindalu-MSFT:
@schjan Thanks for catching this!

PR #368, opened 2018-10-11
Customize manifest from repo

TCourtneyOwen:
This change is dependent on the work the allow customization of the manifest in office-addin-manifest.  This change will of course not be checked in until the office- addin-manifest change is checked in


src/app/index.ts
```diff
@@ -274,7 +275,11 @@ module.exports = yo.extend({
         // Copy project template files from project repository (currently only custom functions has its own separate repo)
         if (projectRepoBranchInfo.repo)
         {
-          git().clone(projectRepoBranchInfo.repo, this.destinationPath(), ['--branch', (projectRepoBranchInfo.branch) ? projectRepoBranchInfo.branch : 'master'], (err) => {
+          git().clone(projectRepoBranchInfo.repo, this.destinationPath(), ['--branch', (projectRepoBranchInfo.branch) ? projectRepoBranchInfo.branch : 'master'], (err) => {        
+            // //modify manifest guid and DisplayName
+            officeAddinManifest.modifyManifestFile(`${this.destinationPath()}/manifest.xml`, 'random', `${this.project.name}`);
```
akrantz: There would be an opportunity here to do the modifyManifestFile in parallel with the .git folder delete. Perhaps the file write time is not that significant, and likely dwarfed by the time to delete the .git folder, but just mentioned that they could be made to be done in parallel.

src/test/new-project.ts
```diff
@@ -361,17 +359,21 @@ describe('Create new project from prompts and command line overrides', () => {
...
 
-      // assert.file(expected);
-      done();
+      assert.file(expected);
+    });
+    
+    it('manifest display name set to customDislayName', async function() {
+      let manifestFilePath = path.win32.resolve('manifest.xml');
+      const info =  await officeAddinManifest.readManifestFile(manifestFilePath);
+      assert.textEqual(info.displayName, customDislayName);
```
akrantz: Also you can verify that a guid was written for the manifest id and that it was changed from the original value.

akrantz:
@TCourtneyOwen, it appears that you merged master into this branch but then closed this PR without merging into master.

src/app/index.ts
akrantz: @TCourtneyOwen You need the callback to be async and await officeAddinManifest.modifyManifestFile()  or use officeAddinManifest.modifyManifestFile().then(() => {
    //rest of the code
}

PR #371, opened 2018-10-18
fix: remove projectInternalName prefix from manifest

### LINKED ISSUE #370: standardize manifest filename to "manifest.xml"
If the manifest file names are standardized to "manifest.xml", then Yo Office would not need to rename them or update package.json script arguments to use the customized name. This would make things simpler.

Usually when the manifest path is used in the developer registry or in the shared folder path, the add-in for the manifest can be distinguished by the directory name in the full path, e.g. "c:\some\path\\**project-name**\manifest.xml".

The only other counter-example I can think of is having other manifests to customize the install location, e..g manifest.xml, manifest.stage.xml, manifest.prod.xml, such that manifest.xml is copied and the urls in the manifest are modified for the alternate install location, but still this doesn't preclude having the basic manifest filename be simply "manifest.xml".

ThorstenHans:
By removing the projectInternalName from the manifest filename, package.json files can be simplified. fixes #370

### DIFF
--- a/src/app/index.ts
+++ b/src/app/index.ts
@@ -31,8 +31,8 @@ module.exports = yo.extend({
   constructor: function () {
     yo.apply(this, arguments);
 
-    this.argument('projectType', { type: String, required: false }); 
-    this.argument('name', { type: String, required: false }); 
+    this.argument('projectType', { type: String, required: false });
+    this.argument('name', { type: String, required: false });
     this.argument('host', { type: String, required: false });
 
     this.option('skip-install', {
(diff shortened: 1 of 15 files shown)


src/app/index.ts
```diff
@@ -1,391 +1,391 @@
...
+            if (fs.existsSync(gitFolder)){
+              helperMethods.deleteFolderRecursively(gitFolder);
+            }
+            return err ? reject(err) : resolve();
+          });
+        }
+        else
+        {
+          /* Copy the manifest */
+          this.fs.copyTpl(this.templatePath(`hosts/${_.toLower(this.project.hostInternalName)}/manifest.xml`), this.destinationPath('manifest.xml'), templateFills);
```
TCourtneyOwen: We will need to update tests is src/tests/manifest-only-projects.ts and src/test/new-projects.ts so that they only look to confirm that "manifest.xml" exists following project generation. Currently they look for the customized manifest file name

ThorstenHans: thanks for that finding

ThorstenHans: @TCourtneyOwen I've updated the tests and had also the chance to remove a couple of unused local vars by moving from the concatenated manifest name to a `const`

ThorstenHans:
> @akrantz I think now that we are doing the async call to copy files, we can uncomment this. This test was failing when we were doing all the file copying/cloning in the context of a synchronous call

I can verify and create a dedicated PR for that one. Otherwise this PR will again be postponed 30 mins due to travis test execution 😄

TCourtneyOwen:
@ThorstenHans regarding enabling the one test, I have a separate PR out that does that so need to make that change. I will be merging this change once I get approval on another PR that's out for a dependent project

PR #378, opened 2018-10-25
Update Yo Office to customize manifest if cloning project

TCourtneyOwen:
- Add new test to validate that the manifest is is modified as part of cloning


akrantz:
Please put the spaces back in the display name to ensure that it works.

akrantz:
Before we merge this, we need to ensure that comments are not removed from the manifest file.

src/test/new-project.ts
```diff
@@ -53,7 +55,7 @@ const expectExcelCustomFunctionFiles = [
   'webpack.config.js',
   'config/web.config',
   'index.html',
-  'src/customfunctions.js',
+  'src/customfunctions.ts',
```
akrantz: File layout has changed in Excel-Custom-Functions.

PR #388, opened 2018-12-10
Fix office-addin-manifest package dependency

akrantz:


akrantz:
@kbrandl What version of Node.js are you running?

kbrandl:
@akrantz I'm running v 6.11.3  (which appears to be pretty old).  What's the minimum required version I should be running for Yo Office to work properly?

akrantz:
I don't know the minimum required, but at least Node 8.x although I am using the latest LTS 10.14.1.

I uninstalled Node and all global packages and reinstalled and verified that Yo Office works now.

akrantz:
Util.promisify was added to Node 8.0 so that would be the minimum version required.

TCourtneyOwen:
Yeah, this is a super-old and we currently run our tests against the stable build 10.14.1.  We should find a way to force the user to use the correct version of node. Maybe we need to update the generator code to detect the version

kbrandl:
Thanks for the feedback, guys. I've upgraded my Node.js to version 10.14.1 (by using the MSI installer on https://nodejs.org/en/) and have version 6.4.1 of **npm**.  Since doing this, I'm now getting the following error whenever I try to install any package (i.e., `npm install...`).  

image

I've also tried completely removing Node.js and reinstalling it (and **npm**) from scratch, but I get the same error whenever I try to install any package. Any suggestions on what I might try to get past this? The error message seems to be saying that my version of **npm** is incompatible with my version of **Node.js** -- but I don't understand how that's possible, given that running the **Node.js** installer should include **npm** along with **Node.js**. What version of **npm** are you guys running? Thanks in advance for any feedback you can provide about this.

kbrandl:
Thanks Courtney. Unfortunately I already tried closing/re-opening the command prompt, but no luck -- same error any time I run `npm install ...`. I even uninstalled Node and npm entirely, rebooted my machine, installed Node and npm from scratch, no luck -- same error any time I run `npm install ...`.  

Just curious -- what version of npm are you running?

TCourtneyOwen:
I am running 6.4.1

akrantz:
@kbrandl Try "npm install -g npm" to upgrade npm to the latest.

kbrandl:
Thanks again, guys. @TCourtneyOwen I'm also running npm 6.4.1 (with Node.js 10.14.1). @akrantz I can't successfully run `npm install -g npm` because any `npm install` I run fails with the same error. 

image

kbrandl:
FWIW, I've searched online and found reports of others getting this error -- but none of the proposed fixes have worked for me. :/

akrantz:
Perhaps go to %APPDATA%\npm and clear out everything in there, then install Node.js again (to populate npm) and then try "npm install -g yo office".

kbrandl:
Thanks for the suggestion, @akrantz. Unfortunately I already tried that, but no luck -- same error any time I run `npm install`.  

Thanks for the ideas, guys...I appreciate your input. Seems like there's just something corrupt about my Node.js / npm install; even uninstalling Node.js and npm, deleting the npm and npm-cache folders from %AppData%, rebooting, and reinstalling Node.js and npm from scratch won't resolve the error.

akrantz:
@kbrandl Another suggestion, try "Node Version Switcher" https://github.com/jasongin/nvs

kbrandl:
@akrantz awesome -- thank you. By using **nvs** to add an older version of node (with the corresponding older version of npm), I was able to successfully run `npm install` to install the new version of generator-office and yo. Thank you!

PR #430, opened 2019-03-20
Pull project templates from repos

TCourtneyOwen:
- Removes host prompting and new-project-test.ts
- Separate PR to follow that will remove all unneccesary files after this change is incorporated

---

    > 
    > * [X ]  Yes
    > * [ ]  No

User will  no longer be prompted for host type and project type strings will change slightly.  Also we now have React Js

    > 
    > * [ X]  Yes
    > * [ ]  No

We need to remove mention of host selection in all documentation. We should also let users know that the manifest and taskpane code now allows for creating an addin for any host type

    Describe manual testing done. 
Created various projects using different project types and verified projects work

4. **Platforms tested**:

    > * [ X] Windows
    > * [X ] Mac


src/app/config/projectProperties.json
```diff
@@ -1,70 +1,58 @@
 {
     "projectTypes": {
-        "angular": {
-            "displayname": "Office Add-in project using Angular framework",
+        "taskpane": {
```
akrantz: I think we should have these specify the "yo-office" branch, just like for Excel-Custom-Functions.

src/app/index.ts
```diff
@@ -6,21 +6,21 @@ import * as _ from 'lodash';
...
 import * as yo from 'yeoman-generator';
-import generateStarterCode from './config/starterCode';
-import projectsJsonData from './config/projectsJsonData';
-import { helperMethods } from './helpers/helperMethods';
-import { modifyManifestFile } from 'office-addin-manifest';
 
-let insight = appInsights.getClient('1ced6a2f-b3b2-4da5-a1b8-746512fbc840');
-let git = require("simple-git");
+const insight = appInsights.getClient('1ced6a2f-b3b2-4da5-a1b8-746512fbc840');
+const git = require("simple-git");
```
akrantz: We need to remove the git requirement and download the zip file instead.

src/app/templates/manifest-only/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "<%= projectInternalName %>",
+  "name": "office-addin-manifest",
```
akrantz: Is this change OK because we use office-addin-manifest to customize the name?

TCourtneyOwen:
I will look into these comments. Any additional comments will be addressed in subsequent versions. Need to wrap this up

kbrandl:
@TCourtneyOwen I thought the outcome of our meeting on March 1 (with @JuaneloJuanelo) was that we weren't moving forward with these changes?  i.e., that Yo Office *should* always prompt the user to select host, and that the resulting project that gets created should only contain content in the manifest and JS file for the selected host. (@JuaneloJuanelo can you please chime in here to make sure everyone's on the same page about this? It's quite possible that I've been out of the loop on conversations that have happened since the meeting we had on March 1.)

TCourtneyOwen:
@kbrandl @akrantz @JuaneloJuanelo @LouMM We don't have the budget to take on the work we discussed on March 1st. For now the plan is to move forward with the removal of the host prompt and for our project templates to work for all hosts.  If the concern is that this will require a lot of documentation updates, I can go in and make these changes myself and send out a PR

src/app/index.ts
TCourtneyOwen: @akrantz Downloading the zip is not as trivial as I hoped. While downloading and extracting the zip is straighforward, the extracted project is created in a subfolder under the project directory, and this will require some clever moving around and deletion of files and folders following the download/unzip.  I am going to open an issue to do this work and will cite the working branch that has these changes

TCourtneyOwen: new issue opened: https://github.com/OfficeDev/generator-office/issues/431

JuaneloJuanelo:
I agree with @kbrandl  that removing the hosts was not settled. I still believe that is important to select the target host, as well as if they want it on JS, TS, Node.

src/app/config/projectProperties.json
TCourtneyOwen: Are you suggesting we create "yo-office" branches for all the projects now?

kbrandl:
@TCourtneyOwen my concern isn't the documentation updates required by these changes to Yo Office, it's that these changes to Yo Office aren't compatible with the developer scenarios we want Yo Office to support and the developer experience we want it to provide. As the DevX PM, @JuaneloJuanelo should provide definitive direction here, but IMHO we'd be better off to *not* ship these changes and to instead wait until we have time/budget to implement the changes we discussed on March 1 (i.e., to always prompt for host and to create a project that only contains XML and JS content for the specified host, implemented in a way that's optimized for efficiency in terms of us maintaining the templates going forward).

src/app/config/projectProperties.json
akrantz: Yes. We don't want merges to master to automatically be deployed into Yo Office.

akrantz: git checkout master
git checkout -b "yo-office"
git push -u origin

TCourtneyOwen: Yeah that makes sense.  I think we should create yo-office branches now then.  Do you agree?

akrantz: Yes

akrantz: We should also make the "yo-office" branch protected in GitHub.

TCourtneyOwen: I created yo-office branches and was able to make most of them protected with the exception of the repos that now no longer have a "Settings" link due to the recent removal of a project admin

src/app/templates/manifest-only/package.json
TCourtneyOwen: reverting this change

TCourtneyOwen:
@kbrandl @JuaneloJuanelo @LouMM @akrantz I created a prototype branch that reintroduces the host prompt to you office.  Currently it works in conjunction with https://github.com/TCourtneyOwen/Office-Addin-TaskPane/tree/add-host-templates which now contains a hosts folder.  

Using https://github.com/TCourtneyOwen/generator-office/tree/readd-host-selection, if you select the Office-Addin-Taskpane project and specify typescript (I only hooked this up for a single project type as a proof-of-concept), yo office will do the following:

- Clone https://github.com/TCourtneyOwen/Office-Addin-TaskPane/tree/add-host-templates
- Move the correct host manifest from hosts/${host} folder to root
- Move the correct source files from hosts/${host}/src to /src
- Delete the hosts directory 

This seems to work well in my preliminary testing and doesn't require to much updating to the Q1 work we did for Yo Office and the various project repos.

I am sure Adam will have thoughts on this, but I think this aligns with what Louis, Juan and I discussed this morning

akrantz:
I do have thoughts on this. It would be better for us to get together and discuss than make random changes. This will have impacts on maintainability and debugging. I doubt these have been fully considered.

akrantz:
I think this pull request can be closed in favor of the newer pull request with this functionality: https://github.com/OfficeDev/generator-office/pull/438

PR #437, opened 2019-04-03
Bundle React by default.

SPDoctor:
Currently, the webpack configuration for the production build marks React and React-DOM as external. This means that out-of-the-box a React add-in works in the development environment but when you go to production build it will result in error that React is undefined, because it is not included in the bundle, nor is it referenced explicitly with script links in the HTML page. This change will result in React being bundled by default. This will add about 3Mb to the bundle, but the client has to load this one way or another. You could explain in the documentation how to omit React from the bundle and use a script link to a CDN to load the React libraries if that is the developer's preference.

---

    > 
    > * [x]  Yes
    > * [ ]  No

config/webpack.prod.js

    > 
    > * [ ]  Yes
    > * [x]  No

    All tests passing. Also manually tested that issue is resolved
4. **Platforms tested**:

    > * [x] Windows
    > * [ ] Mac

### DIFF
--- a/src/app/templates/ts/react/config/webpack.prod.js
+++ b/src/app/templates/ts/react/config/webpack.prod.js
@@ -6,11 +6,6 @@ const ENV = process.env.NODE_ENV = process.env.ENV = 'production';
 module.exports = webpackMerge(commonConfig, {
     devtool: 'source-map',
 
-    externals: {
-        'react': 'React',
-        'react-dom': 'ReactDOM'
-    },
-
     performance: {
         hints: "warning"
     },


TCourtneyOwen:
@akrantz I tested out this PR locally against webpack-dev-server and IIS and everything looks good.  If you are fine with this change, I will go ahead and merge to master,

akrantz:
We're working to move the templates out into separate repositories. The React TypeScript TaskPane template is here: https://github.com/OfficeDev/Office-Addin-TaskPane-React. The webpack config there doesn't have external, so it shouldn't be an issue. @SPDoctor you might want to also check that out and see if it is working well for you.

SPDoctor:
@akrantz Yes, I just cloned that repo and it builds fine for production. I also like the idea of office-adding-dev-certs and the automation of sideloading.

Might be tough to maintain a repo for every case because you have host * language * scenario * framework etc. - might get difficult to make sure any bug fixes or changes are synced across all of them over the long term. I hope you continue to maintain the Yeoman generator as well, which I see as a useful quickstart/demo tool.

akrantz:
We'll still have the Yo Office generator that runs -- we're not getting rid of it. But instead of bundling all of the templates into the generator-office package, only the template that you choose will be downloaded. We can also update templates without having to publish updates of the generator-office package.

We have a better workflow where we can merge changes between the templates.

Glad you like the office-addin-dev-certs! We just completed integrating that into the templates.

SPDoctor:
@akrantz thanks for the explanation. Sounds great :-)

PR #438, opened 2019-04-08
Readd host selection

TCourtneyOwen:

---

    > 
    > * [X ]  Yes
    > * [ ]  No

- We no longer have a 'Jquery' project name. It's now called 'Office-Addin-Taskpane
- We now have a React JS project template
- Users will no longer has to manually register the dev-certs
- The generated project structure will look different (it will look very much like the current Excel Custom Functions structure.

    > 
    > * [X ]  Yes
    > * [ ]  No

- We no longer have a 'Jquery' project name. It's now called 'Office-Addin-Taskpane
- We now have a React JS project template
- Users will no longer has to manually register the dev-certs
- The generated project structure will look different (it will look very much like the current Excel Custom Functions structure.

    Describe manual testing done. 
Generated typescript Office-Addin-Taskpane projects for all hosts on Windows and Mac. Verified projects installed and build successfully. Verified Excel, Word and PowerPoint addin's work since these are the only hosts that currently have source code

4. **Platforms tested**:

    > * [X ] Windows
    > * [ X] Mac


src/app/helpers/helperMethods.ts
```diff
@@ -1,37 +1,123 @@
...
+            {
+                fs.unlinkSync(curPath);
+            }
+        }); 
+        fs.rmdirSync(projectFolder); 
         }
-        else
+    }
+
+    export function doesProjectFolderExists(projectFolder: string)
```
akrantz: nit: remove s from Exists.

src/app/helpers/helperMethods.ts
```diff
@@ -1,37 +1,123 @@
...
-  if (fs.existsSync(projectFolder))
-    {
-      if (fs.readdirSync(projectFolder).length > 0)
-      {          
-        return true;
-      }
+    function _updateMoveProjectFiles(projectFolder: string, host: string, typescript: boolean) {
+        return new Promise(async (resolve, reject) => {
+            // copy host-specific manifest over manifest.xml
+            fs.readFile(path.resolve(`${projectFolder}/manifest.${host}.xml`), 'utf8', (err, contents) => {
```
akrantz: It would be good to have this in a function that specifically is named for what it does, convert the project to a single host. I am not sure if _updateMoveProjectFiles() is a required name or can be renamed, e.g. convertProjectToSingleHost().

akrantz: I would find it simpler to use fs.promise methods so you can write using await rather than the callbacks with the promises. I don't think you need to rewrite it but it would simplify the error handling.

akrantz: I was also thinking that this logic should ideally go into the project so that different projects could have different code to run. This is not required now for the sake of doing it, but it would be better. This code would also need to be skipped for Excel-Custom-Functions[-JS] templates as they are already for a single host.

src/app/helpers/helperMethods.ts
```diff
@@ -1,37 +1,123 @@
...
         }
-    }); 
-    fs.rmdirSync(projectFolder); 
+        return false;
+    };
+
+    export function cleanupProjectFolder(projectFolder: string, host: string, typescript: boolean) {
+        return new Promise(async (resolve, reject) => {
+            try {
+                await _updateMoveProjectFiles(projectFolder, host, typescript);
```
akrantz: If you have two functions, one to move project files and the other to modify package.json, I would put those two into a function named modifyProjectForSingleHost() and have each named accordingly, e.g. modifyProjectFilesForSingleHost() and updatePackageJsonForSingleHost(). Makes it more clear what these functions do. 

Shouldn't there also be a check to skip doing this for Excel-Custom-Functions[-JS]?

src/app/helpers/helperMethods.ts
```diff
@@ -1,37 +1,123 @@
...
+
+                // remove scripts from package.json that are unrelated to selected host,
+                // and update sideload and unload scripts to use selected host.
+                Object.keys(content.scripts).forEach(function (key) {
+                    if (key.includes("sideload:") || key.includes("unload:")) {
+                        delete content.scripts[key];
+                    }
+
+                    if (key == "sideload") {
+                        content.scripts[key] = `office-toolbox sideload -m manifest.xml -a ${host}`
```
akrantz: You should not write out the scripts value here. It should come from the project. You just need to copy the host-specific value to the general value:
```
scripts.[key] = scripts.[`${key}:${host}`];
```

This works for both `sideload` and `unload` scripts so just need switch around it:
```
switch (key) {
   case "sideload":
   case "unload":
     scripts.[key] = scripts.[`${key}:${host}`];
     break;
}
```

src/app/index.ts
```diff
@@ -21,6 +21,7 @@ const excelCustomFunctions = `excel-functions`;
 const manifest = 'manifest';
 const typescript = `Typescript`;
 const javascript = `Javascript`;
+let language = undefined;
```
akrantz: I think you would get a lint error here since you shouldn't need to initialize a variable to `undefined`. `Let language;` is sufficient.

Is tslint being used for generator-office?

src/app/index.ts
```diff
@@ -265,64 +266,41 @@ module.exports = yo.extend({
   {
     return new Promise((resolve, reject) => {
       try {
-        let language = this.project.scriptType === typescript ? 'ts' : 'js';
-        const starterCode = generateStarterCode(this.project.host);
-        const templateFills = Object.assign({}, this.project, starterCode);
+        language = this.project.scriptType === typescript ? 'ts' : 'js';
         let jsonData = new projectsJsonData(this.templatePath());
         let projectRepoBranchInfo = jsonData.getProjectRepoAndBranch(this.project.projectType, language);
+        const templateFills = Object.assign({}, this.project);
```
akrantz: This declaration can be moved into the manifest-only else clause below. It is only used for that as far as I can tell.

src/app/helpers/helperMethods.ts
TCourtneyOwen: Good suggestions. I wasn't aware of fs.promise - good to know!

TCourtneyOwen:
Regarding all the tests we currently have, which will need to be re-written with the new project structure, should we just remove them altogether?  The manifest-only tests are still relevant I guess

src/app/helpers/helperMethods.ts
TCourtneyOwen: Yes, I will add check to skip ECF

src/app/helpers/helperMethods.ts
TCourtneyOwen: I see that when using fs.promises, a warning is displayed: "ExperimentalWarning: The fs.promises API is experimental"

Is there a way to suppress this warning?  It's not something we want users seeing when they run yo office

readme.md
```diff
@@ -58,7 +58,7 @@ $ yo office [arguments] [options]
 The following command line arguments are supported. If using the command line arguments, you must use them in the order cited below, or the generator will prompt you for the values.
 
 #### `projectType`
-Framework to use for the project. The supported project types include JQuery (`jquery`), Angular (`angular`), React (`react`) and Excel Custom Functions (`excel-functions`). You can also use Manifest Only (`manifest`) which will create only the `manifest.xml` for an Office Add-in.
+Framework to use for the project. The supported project types include JQuery (`taskpane`), Angular (`angular`), React (`react`) and Excel Custom Functions (`excel-functions`). You can also use Manifest Only (`manifest`) which will create only the `manifest.xml` for an Office Add-in.
```
kbrandl: Should the start of the second sentence be revised/reworded so that it doesn't use the word `JQuery` ?  (i.e., the options described by the second sentence here should be consistent with the info presented in the Yo Office prompts)

src/app/config/projectProperties.json
```diff
@@ -1,50 +1,67 @@
 {
     "projectTypes": {
-        "angular": {
-            "displayname": "Office Add-in project using Angular framework",
+        "taskpane": {
+            "displayname": "Office Add-in Task Pane project",
```
kbrandl: I notice that we've removed "using JQuery framework" from the end of the name for this project type. Although I understand that this vanilla/basic project type is not technically a `JQuery` project, I propose that we still need some way to differentiate this project type from the other 2 task pane project types (i.e., "...using Angular framework" and "...using React framework").  Technically, each one of the first 3 project types in the list is a "Office Add-in Task Pane project" -- so we should name this first project type in a way that differentiates it from the next two types (Angular and React). 

@JuaneloJuanelo any suggestions for what this basic/vanilla project type should be called?

src/app/index.ts
```diff
@@ -331,17 +302,12 @@ module.exports = yo.extend({
     this.log(`      ${chalk.green('Congratulations!')} Your add-in has been created! Your next steps:\n`);
```
kbrandl: For greater clarity, would be great to show the same type of (colorful) message here at the top of the "post-install hints" as we show when the project creation process begins (changing the word "Creating" to "Created" at the start of the sentence, of course).

image

You could incorporate this message into the existing post-install hints as (roughly) shown here (changing the word "Creating" to "Created" on the second line of text):

image

image

image

image

src/app/index.ts
```diff
@@ -331,17 +302,12 @@ module.exports = yo.extend({
...
     this.log(`         ${chalk.bold('npm start')}\n`);
-    this.log(`      4. Sideload the add-in into your Office application:\n`);
-    this.log(`         ${chalk.bold('npm run sideload')}\n`);
-    this.log(`      5. Open the project in VS Code:\n`);
+    this.log(`      3. Open the project in VS Code:\n`);
     this.log(`         ${chalk.bold('code .')}\n`);
     this.log(`         For more information, visit http://code.visualstudio.com.\n`);
-    this.log(`      Please refer to resource.html in your project for additional information,`);
-    this.log(`      or visit our repo at: https://github.com/officeDev/generator-office.\n`);
+    this.log(`      Please visit https://docs.microsoft.com/en-us/office/dev/add-ins for more information about Office Add-ins.\n`);
```
kbrandl: Please remove the locale (`en-us`) from the docs.microsoft.com URL.  URL should be this:

`https://docs.microsoft.com/office/dev/add-ins`

kbrandl:
In addition to my previous comments, one other thing I'd like to understand: what's the purpose/intent of the `commands` folder (and the `commands.html` file and the `commands.ts` file within that folder) that's present in the project that's created when I choose `Office Add-in Task Pane project` | `TypeScript` | `Excel` ? Should this folder (and the files it contains) be present in the project if I've elected to create a Task Pane project?

akrantz:
The commands folder contains the code related to the add-in commands for the project. https://docs.microsoft.com/en-us/office/dev/add-ins/design/add-in-commands

src/app/index.ts
TCourtneyOwen: fixed

src/app/index.ts
TCourtneyOwen: @kbrandl Not sure I follow - let's discuss offline

src/app/config/projectProperties.json
akrantz: I like "Office Add-in Task Pane". The others are "Office Add-in Task Pane using Angular" or "...using React".

src/app/index.ts
```diff
@@ -331,17 +302,12 @@ module.exports = yo.extend({
     this.log(`      ${chalk.green('Congratulations!')} Your add-in has been created! Your next steps:\n`);
     this.log(`      1. Go the directory where your project was created:\n`);
     this.log(`         ${chalk.bold('cd ' + this._destinationRoot)}\n`);
-    this.log(`      2. Trust the Self-Signed Certificate for your local web server (if you haven't already done that).`);
-    this.log(`         For more information, visit https://github.com/OfficeDev/generator-office/blob/master/src/docs/ssl.md.\n`);
-    this.log(`      3. Start the local web server:\n`);
+    this.log(`      2. Start the local web server and sideload the add-in:\n`);
```
akrantz: It would be better to replace step 3 with step 2 (open in VS Code), and then have the next step be to use `Terminal, Run Task, "Dev Server"` to run the local web server in VS Code. Instead of sideload, "Start without Debugging" can be used to launch the host and load the add-in.

akrantz: Note that for Excel Custom Functions, "Start Debugging" should be used instead of "Start Without Debugging" because you can debug the custom functions runtime with VS Code. 

@kbrandl Perhaps we should provide a link to docs that explain how to do these steps. I think there is value in also providing some steps here as some users may not click the link.

akrantz: We can also merge these changes, and then continue to refince this afterwards...

src/app/config/projectProperties.json
kbrandl: Upon further consideration, I think it's fine as is ("Office Add-in Task Pane project").

TCourtneyOwen: I will update the string when I get into the office

TCourtneyOwen: I agree with Kim. Let's leave as is

readme.md
akrantz: @TCourtneyOwen So what about the usage of "JQuery"? I don't think that word should be used.

src/app/helpers/helperMethods.ts
```diff
@@ -1,37 +1,146 @@
...
+            }
+        }); 
+        fs.rmdirSync(projectFolder); 
         }
-        else
+    }
+
+    export function doesProjectFolderExist(projectFolder: string)
+    {      
+    if (fs.existsSync(projectFolder))
```
akrantz: nit: indentation in this function

src/app/helpers/helperMethods.ts
```diff
@@ -1,37 +1,146 @@
...
+        try {
+            // copy host-specific manifest over manifest.xml
+            const manifestContent: any = await readFileAsync(path.resolve(`${projectFolder}/manifest.${host}.xml`), 'utf8');
+            await writeFileAsync(path.resolve(`${projectFolder}/manifest.xml`), manifestContent);
+
+            switch (projectType) {
+                case "taskpane":
+                {
+                    // copy host-specific taskpane.ts[js] over src/taskpane/taskpane.ts[js]
+                    const srcContent = await readFileAsync(path.resolve(`${projectFolder}/src/taskpane/${host}.${typescript ? 'ts' : 'js'}`), 'utf8');
```
akrantz: You could use: `const extension = typescript ? "ts" : "js";` and then use extension in relevant places.

TCourtneyOwen:
@akrantz , yeah we should definitely not use jQuery anywhere in here. Need to find a suitable word to replace it

readme.md
kbrandl: Would "Vanilla JS" be an accurate/appropriate way of describing that project type (instead of "JQuery")?

src/app/index.ts
kbrandl: @TCourtneyOwen This isn't critical feedback; I'm going to mark this conversation as Resolved.

readme.md
akrantz: I don't think so. I would change the text to:

+ Specify the project type to create. `taskpane` for a TaskPane add-in using HTML, `angular` for a TaskPane add-in using the Angular framework, `react` for a TaskPane add-in using the React framework, `excel-functions` for a Task Pane add-in with Excel Custom Functions, or `manifest` for just the manifest and related files needed for an Office Add-in.

akrantz: @TCourtneyOwen I pushed a commit to use a table to show the values.

src/app/helpers/helperMethods.ts
TCourtneyOwen: fixed locally

src/app/helpers/helperMethods.ts
TCourtneyOwen: good idea

PR #483, opened 2019-08-28
Add usage data package

TCourtneyOwen:

---

    > 
    > * [X ]  Yes
    > * [ ]  No

We now give a prompt asking if user wants to opt out of gather usage data

    > 
    > * [x ]  Yes
    > * [ ]  No

We updated the ReadMe to add a privacy statement with a pointer to a more specific usage-data.md in Office-Addin-Scripts

    Describe manual testing done. 
Created projects to ensure prompting works as expected and that usage data is properly sent to applicaition insighjs

4. **Platforms tested**:

    > * [x ] Windows
    > * [ ] Mac


src/app/index.ts
```diff
@@ -160,37 +196,40 @@ module.exports = yo.extend({
...
+        ScriptType: [this.project.scriptType],
+        IsManifestOnly: [this.project.isManifestOnly.toString()],
+        ProjectType: [this.project.projectType, durationForProjectType],
+        IsProdData: [isProdData]
+      };
+      // Send usage data for project created
+      usageDataObject.reportEvent(defaults.promptSelectionstEventName, projectInfo);
     } catch (err) {
-      insight.trackException(new Error('Prompting Error: ' + err));
+      usageDataObject.reportError(isProdData ? defaults.promptSelectionsErrorEventName : defaults.testPromptSelectionsErrorEventName, new Error('Prompting Error: ' + err));
```
akrantz: You might consider having reportError() be able to append "-test" to the event name when isForTesting is true. You wouldn't have to make different event names for every event and write code to test it on the call. Rather the function implementation can know and do it automatically.

src/app/index.ts
```diff
@@ -68,6 +75,12 @@ module.exports = yo.extend({
       desc: 'Use the prerelease version of the project template.'
     });
 
+    this.option('is-test', {
```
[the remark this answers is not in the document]
akrantz: I said this before but continer "--testing" (or "--for-testing" or "--test") for the name. You don't really need "is" since specifying a flag indicates true by virtue of it being present (and false if not specified).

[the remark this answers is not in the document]
TCourtneyOwen: changed to --test

src/app/index.ts
TCourtneyOwen: great idea

src/app/index.ts
```diff
@@ -1,32 +1,39 @@
...
-import { helperMethods } from './helpers/helperMethods';
-import { modifyManifestFile } from 'office-addin-manifest';
 
-let insight = appInsights.getClient('1ced6a2f-b3b2-4da5-a1b8-746512fbc840');
 const childProcessExec = promisify(childProcess.exec);
 const excelCustomFunctions = `excel-functions`;
 const manifest = 'manifest';
 const typescript = `TypeScript`;
 const javascript = `JavaScript`;
+let isProdData: boolean = true;
```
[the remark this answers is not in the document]
TCourtneyOwen: Yeah some stuff needs to be global due to callbacks

src/app/index.ts
```diff
@@ -160,37 +196,40 @@ module.exports = yo.extend({
...
-    })
-    .catch((err) => {
-      insight.trackException(new Error('Installation Error: ' + err));
-      process.exitCode = 1;
-    });
+      .then(() => {
+        done();
+      })
+      .catch((err) => {
+        usageDataObject.reportError(`${defaults.copyFilesErrorEventName}${isForTesting ? "-test" : ""}`, new Error('Installation Error: ' + err));
```
akrantz: In case it wasn't clear, I meant to not add "-test" here by the caller, but rather modify the implementation of reportError so that when options.isForTesting is true, it would append "-test". The advantage is that it will be consistently done. It does mean having to make the change in the package, but I think it would be worth doing that.

I presume the desire to use a different event name is because it makes it easier to query for the non "test" events, rather than query all the events where the property value "isForTesting" is false. 

I do think it would be better to have the "isForTesting" option as part of the options passed in, and have the implementation add this to the event data properties automatically. Then you should see the "-test" events have property "isForTesting = true" and the other events should always have isForTesting  = false.

src/app/index.ts
```diff
@@ -94,8 +94,8 @@ module.exports = yo.extend({
     if (this.options.details) {
       this._detailedHelp();
     }
-    if (this.options['is-test']) {
-      isProdData = false;
+    if (this.options['test']) {
+      isForTesting = false;
```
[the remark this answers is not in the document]
TCourtneyOwen: Good catch.

src/app/index.ts
TCourtneyOwen: I like this suggestion and can make a change in office-addin-usage-data, but this will require that this change pushed and published almost immediately since I will have to bump the version of office-addin-usage-data in generator-office to pick up this change.  Might be worth holding off incorporating this change into this generator-office PR

akrantz: My view is that it would be better to make the update in office-addin-scripts and publish the updated package now, rather than try to make this change later.

TCourtneyOwen: OK, then this has to get done today - particularly with you being out of town for the next week.  We can interminably have the train idling in the station and things will begin to snowball and next thing we know it's two weeks later and we still haven't wrapped this up.

PR #511, opened 2020-01-22
Add Outlook host for Single-Sign-On

TCourtneyOwen:
- Make supported hosts data-driven by projectProperties.json

---

    > 
    > * [x ]  Yes
    > * [ ]  No

This adds Outlook as a suggest host for single-sign-on. This change shouldn't be merged until we make the necessary changes to the Office-Addin-Taskpane-SSO templates and to the office-addin-sso package (separate PRs coming out)

    > 
    > * [x ]  Yes
    > * [ ]  No

We will need to update SSO documentation to say that Outlook is supported

    Describe manual testing done. 
Confirmed new hosts population works
4. **Platforms tested**:

    > * [x ] Windows
    > * [x] Mac

### DIFF
--- a/src/app/config/projectsJsonData.ts
+++ b/src/app/config/projectsJsonData.ts
@@ -49,15 +49,11 @@ export default class projectsJsonData {
     return this.m_projectJsonData.projectTypes[_.toLower(projectType)].templates.javascript != undefined && this.m_projectJsonData.projectTypes[_.toLower(projectType)].templates.typescript != undefined;
   }
 
-  getHostTemplateNames(isSsoProject: boolean) {
+  getHostTemplateNames(projectType: string) {
     let hosts: string[] = [];
-    for (let key in this.m_projectJsonData.hostTypes) {
-      if (isSsoProject) {
-        if (key === 'excel' || key === 'word' || key === 'powerpoint') {
-          hosts.push(this.m_projectJsonData.hostTypes[key].displayname);
-        }
-      } else {
-        hosts.push(this.m_projectJsonData.hostTypes[key].displayname);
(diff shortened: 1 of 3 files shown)


src/app/config/projectsJsonData.ts
```diff
@@ -49,15 +49,11 @@ export default class projectsJsonData {
...
+  getHostTemplateNames(projectType: string) {
     let hosts: string[] = [];
-    for (let key in this.m_projectJsonData.hostTypes) {
-      if (isSsoProject) {
-        if (key === 'excel' || key === 'word' || key === 'powerpoint') {
-          hosts.push(this.m_projectJsonData.hostTypes[key].displayname);
-        }
-      } else {
-        hosts.push(this.m_projectJsonData.hostTypes[key].displayname);
+    for (let key in this.m_projectJsonData.projectTypes) {
```
akrantz: This isn't a change that you need to make but just saw it and wanted to comment.

TypeScript added an equivalent of the null-coalescing operator in C# (?.) which returns undefined if null or undefined. So it might be possible to change implementations such as this loop to be a more direct call:

```
const hosts: string[] = this.m_projectJsonData.projectTypes[key]?.projectTypes ?? [];
```

So if the mapping for the key returned undefined, it falls back to an empty array, and if defined, returns the array. This would avoid having to actually loop through, so O(1) instead of O(n).

```

TCourtneyOwen: Not sure what's causing the failure on the Windows CI run. In fact the CI loop bails out immediately without testing anything

TCourtneyOwen: CI loop passing now - must've been some intermittent issue with the CI agent machine

PR #549, opened 2020-07-29
Fixes issue where postInstallHints wasn't getting called

TCourtneyOwen:
- The postInstallHints method was no longer getting called due to an apparent change in the way callbacks are handled by yeoman-generator.  This issue presumably surfaced when we upgraded the yeoman-generator version a few months back.  What apparently happens is that the postInstallHints callback is simply ignored following install
- We believe this may be a source of confusion for the SSO template in particular where we give explicit instructions to run "configure-sso" prior to running "npm start"
- The fix is to call postInstallHints in the yeoman-generator "end" method, which is "Called last, cleanup, say good bye, etc"
- Ignoring the call to postInstallHints for test code, since postInstallHints also calls process.exit(), which will cause tests to bail out
- Also update package-lock.json to address vulnerabilities

---

    > 
    > * [ ]  Yes
    > * [x ]  No

    > 
    > * [ ]  Yes
    > * [x ]  No

    Did manual test to ensure postInstallHints are now displayed at the end of project generation.  Ran tests locally to ensure they all pass

4. **Platforms tested**:

    > * [ x] Windows
    > * [ ] Mac

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3041,9 +3041,9 @@
       }
     },
     "lodash": {
-      "version": "4.17.15",
-      "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.15.tgz",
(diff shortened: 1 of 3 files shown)


akrantz:
@TCourtneyOwen This pull request is still active. Your latest release notes implied that it was included.

PR #573, opened 2020-11-30
Dependabot Addition

arttarawork:

---

    > 
    > * [ ]  Yes
    > * [x]  No

Helps us upkeep dependencies for security.

    > 
    > * [ ]  Yes
    > * [x]  No

    Describe manual testing done. 

`npm i && npm run build && npm run test`

All 15 tests pass locally

4. **Platforms tested**:

    > * [x] Windows
    > * [ ] Mac

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,3 +9,6 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+      time: "08:00"
+      timezone: "America/Los_Angeles"
+    open-pull-requests-limit: 5


.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
```
TCourtneyOwen, line 9: I see there are additional settings we can add such as the following:

  schedule:
    interval: daily
    time: "13:00"
  open-pull-requests-limit: 10

Should we set a PR limit?  I fear this could result in just too much noise for us to deal with. Also we should probably set the "time" value to a time when out VMs are online.  They are offline for deployment from 1:00 AM - ~5:00 AM PST every day and we don't want PRs getting created then because the CI jobs will fail if the VMs are unavailable

TCourtneyOwen:
Let's go ahead and merge this as a proof-of-concept.  We will see how (and if) this works. I see there are also apparently ways to do auto-merge on patch builds, but I don't want to enable that yet until we are more familiar with how this all works.  Based on the .yml setting, we should get a max of five PRs per week for this repo. We will see how manageable this is. I fear that if we enable dependabot across all our repos, we will just get too many PRs and this won't be sustainable, but at least if we limit the number of PRs we may get good bang for the buck (i.e. some but not all our packages will be updated) and still have this be a sustainable approach

TCourtneyOwen (approved):

TCourtneyOwen:
So this PR limit is not working as I would expect. After merging this change, I immediately see more than five PRs get generated. While I am happy to see the dependabot kick off, this is just too many PRs I think.  Also, I hope the weekly setting works because I don't want to see these PRs get opened every day

TCourtneyOwen:
Actually, maybe it is only five PRs, but two mails are getting sent for each PR, one specifying all the changes contained in the package update and the other containing a link to the PR

PR #610, opened 2021-02-23
Creating new cpy command

igor-ribeiiro:
CPX isn't maintained and currently has a vulnerability issue. So it is being changed to CPY, which is already used in other repositories within this organization

---

    > 
    > * [ ]  Yes
    > * [X]  No

    > 
    > * [ ]  Yes
    > * [X]  No

Tested tunning the command and seeing that the files created are exactly the ones created before

4. **Platforms tested**:

    > * [X] Windows
    > * [ ] Mac

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,9 +7,7 @@
     "url": "https://github.com/OfficeDev/generator-office/"
   },
   "scripts": {
-    "copy": "cpx \"src/app/templates/**/*.*\" generators/app/templates && cpx \"src/app/config/**/projectProperties.json*\" generators/app/templates && cpx \"src/app/config/**/projectProperties.json*\" generators/test",
-    "copy-watch": "cpx \"src/app/templates/**/*.*\" generators/app/templates --watch",
-    "copy2": "cpy templates/**/*.* ../../generators/app --parents --cwd=src/app && cpy ./**/projectProperties.json* ../../../generators/app/templates --parents --cwd=src/app/config && cpy ./**/projectProperties.json* ../../../generators/test --parents --cwd=src/app/config",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -7,8 +7,7 @@
     "url": "https://github.com/OfficeDev/generator-office/"
   },
   "scripts": {
-    "copy": "cpx \"src/app/templates/**/*.*\" generators/app/templates && cpx \"src/app/config/**/projectProperties.json*\" generators/app/templates && cpx \"src/app/config/**/projectProperties.json*\" generators/test",
-    "copy-watch": "cpx \"src/app/templates/**/*.*\" generators/app/templates --watch",
+    "copy": "cpy templates/**/*.* ../../generators/app --parents --cwd=src/app && cpy ./**/projectProperties.json* ../../../generators/app/templates --parents --cwd=src/app/config && cpy ./**/projectProperties.json* ../../../generators/test --parents --cwd=src/app/config",
```
akrantz, line 10: Do you have to use the `--cwd` command, or could the paths just be relative to the project directory as before?

akrantz (commented):
This looks fine but I am wondering if the --cwd parameter is needed.

package.json
igor-ribeiiro: Yeah, it is needed to preserve the same file structure as using the previous command.
If we don't use it, it would generate a file structure including all the paths until the file. Something like:
```src/app/templates/...``` instead of the correct: ```templates...```

TCourtneyOwen: Does changing the package to cpy in fact remove any remaining vulnerabilities?

TCourtneyOwen: We should also be sure that a project generated with these changes in fact works as expected. I think the manifest-only type is the one most likely to be impacted as we copy over the files from the templates dir to the buit generators dir

igor-ribeiiro: Yes, the vulnerabilities were fixed.
Before the change, there were: ```5 low severity vulnerabilities```
Now there are: ```0 vulnerabilities```

akrantz (approved):

TCourtneyOwen (approved):

PR #693, opened 2022-03-16
Json template

igor-ribeiiro:
Creating an option to create a new JSON template for yo office. Merging in the "json" branch for now in order to not be available in master for now.

---

    > 
    > * [X]  Yes
    > * [ ]  No

There is a new preview option for the generator office. Currently is in just a branch, but when merged with master it will impact the user experience

    > 
    > * [ ]  Yes
    > * [X]  No

    Not right now, but there will need more changes in the documentation in the future.

    Linked and test the happy scenario of choosing the preview option.

4. **Platforms tested**:

    > * [X] Windows
    > * [ ] Mac

### DIFF
--- a/src/app/config/projectProperties.json
+++ b/src/app/config/projectProperties.json
@@ -126,11 +126,6 @@
         "json-based-manifest": {
             "displayname": "[Preview] Json Based Add-in Project",
             "templates": {
-                "javascript": {
-                    "repository": "https://github.com/OfficeDev/Office-Addin-TaskPane-JS",
-                    "branch": "yo-office",
-                    "prerelease": "yo-office-prerelease"
-                },
                 "typescript": {
                     "repository": "https://github.com/OfficeDev/Office-Addin-TaskPane",
                     "branch": "yo-office",
--- a/src/app/config/projectProperties.json
+++ b/src/app/config/projectProperties.json
@@ -122,6 +122,24 @@
             "supportedHosts": [
                 "Excel"
             ]
+        },
+        "json-based-manifest": {
+            "displayname": "[Preview] Json Based Add-in Project",
+            "templates": {
+                "javascript": {
+                    "repository": "https://github.com/OfficeDev/Office-Addin-TaskPane-JS",
+                    "branch": "yo-office",
+                    "prerelease": "yo-office-prerelease"
+                },
+                "typescript": {
+                    "repository": "https://github.com/OfficeDev/Office-Addin-TaskPane",
+                    "branch": "yo-office",
+                    "prerelease": "yo-office-prerelease"
+                }
+            },
+            "supportedHosts": [
+                "Outlook"
+            ]
         }
     },
     "hostTypes": {
--- a/src/app/config/projectProperties.json
+++ b/src/app/config/projectProperties.json
@@ -128,8 +128,8 @@
             "templates": {
                 "typescript": {
                     "repository": "https://github.com/OfficeDev/Office-Addin-TaskPane",
-                    "branch": "yo-office",
-                    "prerelease": "yo-office-prerelease"
+                    "branch": "json",
+                    "prerelease": "json"
                 }
             },
             "supportedHosts": [
--- a/src/app/config/projectProperties.json
+++ b/src/app/config/projectProperties.json
@@ -128,8 +128,8 @@
             "templates": {
                 "typescript": {
                     "repository": "https://github.com/OfficeDev/Office-Addin-TaskPane",
-                    "branch": "json",
-                    "prerelease": "json"
+                    "branch": "json-yo-office",
+                    "prerelease": "json-yo-office-prerelease"
                 }
             },
             "supportedHosts": [
--- a/src/app/config/projectsJsonData.ts
+++ b/src/app/config/projectsJsonData.ts
@@ -59,6 +59,21 @@ export default class projectsJsonData {
     return hosts;
   }
 
+  getSupportedScripts(projectType: string) {
+    const scripts: string[] = [];
+    for (const template in this.m_projectJsonData.projectTypes[projectType].templates) {
+      let script: string;
+      if (template === "javascript") {
+        script = "JavaScript";
+      } else if (template === "typescript") {
+        script = "TypeScript";
+      }
+
+      scripts.push(script);
+    }
+    return scripts;
+  }
+
   getHostDisplayName(hostKey: string) {
     for (const key in this.m_projectJsonData.hostTypes) {
       if (_.toLower(hostKey) == key) {
--- a/src/app/config/projectsJsonData.ts
+++ b/src/app/config/projectsJsonData.ts
@@ -59,19 +59,19 @@ export default class projectsJsonData {
     return hosts;
   }
 
-  getSupportedScripts(projectType: string) {
-    const scripts: string[] = [];
+  getSupportedScriptTypes(projectType: string) {
+    const scriptTypes: string[] = [];
     for (const template in this.m_projectJsonData.projectTypes[projectType].templates) {
-      let script: string;
+      let scriptType: string;
       if (template === "javascript") {
-        script = "JavaScript";
+        scriptType = "JavaScript";
       } else if (template === "typescript") {
-        script = "TypeScript";
+        scriptType = "TypeScript";
       }
 
-      scripts.push(script);
+      scriptTypes.push(scriptType);
     }
-    return scripts;
+    return scriptTypes;
   }
 
   getHostDisplayName(hostKey: string) {
--- a/src/app/config/projectsJsonData.ts
+++ b/src/app/config/projectsJsonData.ts
@@ -100,16 +100,13 @@ export default class projectsJsonData {
     return undefined;
   }
 
-  getProjectRepoAndBranch(projectTypeKey: string, scriptType: string, prerelease: boolean, isJsonProject: boolean) {
+  getProjectRepoAndBranch(projectTypeKey: string, scriptType: string, prerelease: boolean) {
     scriptType = scriptType === 'ts' ? 'typescript' : 'javascript';
     const repoBranchInfo = { repo: <string>null, branch: <string>null };
 
     repoBranchInfo.repo = this.getProjectTemplateRepository(projectTypeKey, scriptType);
-    if (isJsonProject) {
-      repoBranchInfo.branch = "json";
-    } else {
-      repoBranchInfo.branch = (repoBranchInfo.repo) ? this.getProjectTemplateBranchName(projectTypeKey, scriptType, prerelease) : undefined;
-    }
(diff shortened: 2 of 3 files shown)


src/app/config/projectProperties.json
```diff
@@ -122,6 +122,19 @@
             "supportedHosts": [
                 "Excel"
             ]
+        },
+        "json-based-manifest": {
+            "displayname": "[Preview] Json Based Add-in Project",
+            "templates": {
+                "typescript": {
+                    "repository": "https://github.com/OfficeDev/Office-Addin-TaskPane",
```
millerds, line 130: I think this has to point to the branch in the repo doesn't it?

src/app/config/projectsJsonData.ts
```diff
@@ -100,13 +100,16 @@ export default class projectsJsonData {
     return undefined;
   }
 
-  getProjectRepoAndBranch(projectTypeKey: string, scriptType: string, prerelease: boolean) {
+  getProjectRepoAndBranch(projectTypeKey: string, scriptType: string, prerelease: boolean, isJsonProject: boolean) {
```
millerds, line 103: If you use the right branch URL in the ProjectProperties.json then you shouldn't have to use a different boolean here to override the branch name.

src/app/index.ts
```diff
@@ -23,10 +23,12 @@ const yo = require("yeoman-generator"); // eslint-disable-line @typescript-eslin
 const childProcessExec = promisify(childProcess.exec);
 const excelCustomFunctions = `excel-functions`;
 let isSsoProject = false;
+let isJsonProject = false;
```
millerds, line 26: I don't like the pattern I'm seeing where there is a new is*Project and project name variable being created for each special case.  This is not a good pattern or precedent that we should continue using.

millerds (changes requested):

src/app/config/projectProperties.json
igor-ribeiiro: The problem is that the code already has some functions to use specific branches.
So I was modifying these functions instead of putting the branch in the URL, as if I just put the branch in the URL, it wouldn't be enough (as the code would be adding even more branch information).

src/app/config/projectsJsonData.ts
igor-ribeiiro: Then I would have a condition for this function to not get called. As this functions returns something that will get appended to the URL. So we cannot just put the URL as it is without some scape path for the JSON condition

src/app/index.ts
igor-ribeiiro: I will try to use `manifestType` then. I think I already started but I decided to continue with the pattern I was seeing on the file.

millerds:
Supposedly at one point the shared host branch in custom functions was referenced here in generator-office.  I'm not sure how that was accomplished then, but that would be a similar situation to this.

src/app/config/projectProperties.json
millerds: What parts of the code are looking at specific branches?

millerds: OK . . . I think I get it now.  I do think you want the repository to be the json template branch, but the "branch" and "prerelease" should not be the standard branches used in other property sets.  If you look at the CF template it's has a separate set of branches for the code and for the yo office and for the prerelease.

src/app/config/projectsJsonData.ts
millerds: This is returning information about the template repo and branches.  I don't see where it's getting appended, but it is getting it's properties looked up.

src/app/index.ts
millerds: I'm not sure why the manifestType is needed as well.  generator-office should have to know details of the contents of a template.

src/app/config/projectsJsonData.ts
igor-ribeiiro: This is where the branch gets appended and the URL gets changed:
```
    export async function downloadProjectTemplateZipFile(projectFolder: string, projectRepo: string, projectBranch: string): Promise<void> {
        const projectTemplateZipFile = `${projectRepo}/archive/${projectBranch}.zip`;
```
This is in `src\app\helpers\helperMethods.ts`. As we can see, the URL `projectRepo` gets changed and it not what is actually used.

src/app/config/projectProperties.json
igor-ribeiiro: Yeah, I understand what you mean. I will try to change those values and see if I can get it to work

src/app/index.ts
igor-ribeiiro: I need some variables in order to reduce the number of questions asked. CFs has specific ones to say to only use excel for example.
I need someway to make it so we don't need questions for Host (only outlook) and for ts vs js (only typescript).

millerds: Use the size of the supportedHosts array to determine if that question needs to be asked.

This is the first time we've had a case where we aren't saying ts v. js.  We could make a branch off of the js version of the template repo and keep the question, or we could add that to the properties json and use that (in a similar way to the supportedHosts).  Seems like the code should be generic and the differences should be in the properties json.

src/app/config/projectsJsonData.ts
millerds: I see.  So yes, the base repo should be the base url w/out the branch . . . but the branches should point to yo-office and yo-ffice-prerlease branches for the new template.

igor-ribeiiro: I just changed the branch names in the `.json` config file to the new `json` branch.
How can I make the branches point to `yo-office` as it is not a new repository? We need to make it link to the associated json branch

src/app/index.ts
igor-ribeiiro: I will change it so we still only have one template (the typescript one) as it will be easier to iterate. We can quickly change if needed later.
I will add that in the next commit

src/app/config/projectsJsonData.ts
millerds: Looks at the CF repo for the example (shared app branches).  The core template is the root + yo-office + yo-office-prerelease branches, and the new template uses the json + json-yo-office + json-yo-office-prerelease branches.  Generator office doesn't reference the root branch directly, just the other two branches designated in the properties json file.  This is how we control what is in use vs still in "development".  Updating these is part of the repo merge process I go through.

igor-ribeiiro: I just created those 2 braches and updates the code to direct to them

src/app/index.ts
millerds: We should check for property array sizes and set the value automatically if they are just one instead of making separate variable to track each think.  So something like (this is not exact) "if (jsonData[answerForProjectType.projectType][templates].size == 1) answerForScriptType.scriptType = jsonData[answerForProjectType.projectType][templates][0];" can be used to set the value without asking the question.

igor-ribeiiro:
This branch is ready to be revied again 😄 but github doesn't let me "request" again, so I'm sending this message @millerds

src/app/config/projectsJsonData.ts
```diff
@@ -59,6 +59,21 @@ export default class projectsJsonData {
     return hosts;
   }
 
+  getSupportedScripts(projectType: string) {
```
millerds, line 62: "Scripts" is confusing.  I see that this is for the "scriptType" questions and while I'm not a fan of that labeling either, I think this should match.

src/app/index.ts
```diff
@@ -185,12 +188,15 @@ module.exports = class extends yo {
           name: 'scriptType',
           type: 'list',
           message: 'Choose a script type:',
-          choices: [typescript, javascript],
-          default: typescript,
-          when: !this.options.js && !this.options.ts && !isManifestProject
+          choices: jsonData.getSupportedScripts(projectType),
```
millerds, line 191: Should get the collection once and the reference it in each of the necessary places.

src/app/index.ts
```diff
@@ -209,17 +215,18 @@ module.exports = class extends yo {
         name: 'host',
         message: 'Which Office client application would you like to support?',
         type: 'list',
-        default: 'Excel',
-        choices: jsonData.getHostTemplateNames(answerForProjectType.projectType).map(host => ({ name: host, value: host })),
+        default: jsonData.getHostTemplateNames(projectType)[0],
+        choices: jsonData.getHostTemplateNames(projectType).map(host => ({ name: host, value: host })),
         when: (this.options.host == null || this.options.host != null && !jsonData.isValidInput(this.options.host, true /* isHostParam */))
-          && !isExcelFunctionsProject
+          && !isExcelFunctionsProject && jsonData.getHostTemplateNames(projectType).length > 1
```
millerds, line 221: Do we need the 'isExcelFunctionsProject' anymore?

src/app/index.ts
```diff
@@ -233,6 +240,7 @@ module.exports = class extends yo {
       usageDataObject.reportEvent(defaults.promptSelectionstEventName, projectInfo);
     } catch (err) {
       usageDataObject.reportError(defaults.promptSelectionsErrorEventName, new Error('Prompting Error: ' + err));
+      throw err;
```
millerds, line 243: What's the side effect of throwing the error when we didn't before (we were eating the error)?

millerds (changes requested):

src/app/index.ts
igor-ribeiiro: Yo office would continue the process saying it was going to download an `undefined` project for host `undefined`.
That download would fail and wouldn't throw anything meaningful. So I thought we should thrown the error

millerds: If that was the experience before, then what's the experience when throwing the error?

igor-ribeiiro: It would simply throw the error for the client and terminate the process without trying to download and fail then.
The error thrown is any error message we can write to customers. The way it was, customer wouldn't see any errors we wrote

igor-ribeiiro:
Ready for review again 😄

src/app/index.ts
igor-ribeiiro: Sure, just removed this line

src/app/index.ts
```diff
@@ -180,17 +183,21 @@ module.exports = class extends yo {
...
       const askForScriptType = [
         {
           name: 'scriptType',
           type: 'list',
           message: 'Choose a script type:',
-          choices: [typescript, javascript],
-          default: typescript,
-          when: !this.options.js && !this.options.ts && !isManifestProject
+          choices: getSupportedScriptTypes,
+          default: getSupportedScriptTypes,
```
millerds, line 193: Shouldn't this be just one type?

millerds (changes requested):

src/app/index.ts
igor-ribeiiro: Nice catch, it should be

millerds (approved):

PR #706, opened 2022-05-05
Fixing install not working on json manifests

igor-ribeiiro:
I have to update package versions because they were outdated to use some packages that did not handle JSON manifests yet.

---

    > 
    > * [X]  Yes
    > * [ ]  No

Changed the `name` that appears when the project is created from `json-based-manifest` to `teams-manifest`.
Made `npm install` work after using a json manifest.

    > 
    > * [ ]  Yes
    > * [X]  No

    Tested creating `.xml` and `.json` projects and noticed both ran until the end with all steps passings.

4. **Platforms tested**:

    > * [X] Windows
    > * [ ] Mac


package.json
```diff
@@ -41,34 +41,34 @@
...
-    "mocha": "^9.2.0",
-    "prettier": "^2.5.1",
-    "ts-node": "^9.0.0",
-    "typescript": "^3.9.2",
+    "eslint": "^8.14.0",
+    "eslint-config-prettier": "8.5.0",
+    "mocha": "^10.0.0",
+    "prettier": "^2.6.2",
+    "ts-node": "^10.7.0",
+    "typescript": "^4.6.4",
```
millerds, line 69: Be careful about updating typescript versions.  I can't remember where it was, but once when I tried this it has some unintended consequences and I had to revert back to a previous version.

millerds (commented):
I feel like the manifest type should be in the json file like all of the other unique template settings are.  We should avoid special casing templates in the code itself (I know there are things like that already, but we shouldn't make the problem worse).

package.json
igor-ribeiiro: It didn't seem to break anything in yo-office. I will double-check the tests but will process updating if they are all successful.

src/app/config/projectProperties.json
```diff
@@ -2,6 +2,7 @@
     "projectTypes": {
         "taskpane": {
             "displayname": "Office Add-in Task Pane project",
+            "manifestType": "xml",
```
millerds, line 5: Instead of specifying the manifest type you should specify the manifest relative path.  That way is a single lookup without any if statements that have special information about subfolders for specific project types.

millerds (changes requested):

millerds (approved):

PR #714, opened 2022-06-13
Fixing ts option

igor-ribeiiro:
This PR makes the `--ts` option work again and makes typescript be chosen. It had a bug that the option was getting ignored and the generated project was javascript.

---

    > 
    > * [ ]  Yes
    > * [x]  No

    > 
    > * [ ]  Yes
    > * [x]  No

Tested that the `yo office --ts` options are now creating a typescript project.

4. **Platforms tested**:

    > * [X] Windows
    > * [ ] Mac

### DIFF
--- a/src/app/index.ts
+++ b/src/app/index.ts
@@ -196,9 +196,6 @@ module.exports = class extends yo {
         }
       ];
       const answerForScriptType = await this.prompt(askForScriptType);
-      if (!answerForScriptType.scriptType && !this.options.js && !this.options.ts) {
-        answerForScriptType.scriptType = getSupportedScriptTypes[0];
-      }
 
       /* askforName will be triggered if no project name was specified via command line Name argument */
       const askForName = [{
@@ -223,9 +220,6 @@ module.exports = class extends yo {
           && jsonData.getHostTemplateNames(projectType).length > 1
       }];
       const answerForHost = await this.prompt(askForHost);
-      if (!answerForHost.host && !this.options.host) {
-        answerForHost.host = jsonData.getHostTemplateNames(projectType)[0];
-      }
       const endForHost = (new Date()).getTime();
       const durationForHost = (endForHost - startForHost) / 1000;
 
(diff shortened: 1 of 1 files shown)


millerds:
So how does this change actually fix the problem?

igor-ribeiiro:
> So how does this change actually fix the problem?


In: https://github.com/OfficeDev/generator-office/blob/master/src/app/index.ts#L302 the `answerForScriptType.scriptType` is true even if no answer was provided.

This bug was introduced when we wanted to make the `preview` template option in Yo Office, as it didn't have a language and we didn't want to ask the language question. At that time, we didn't think about the `--ts` option and its impact on it.

millerds:
> > So how does this change actually fix the problem?
> 
> In: https://github.com/OfficeDev/generator-office/blob/master/src/app/index.ts#L302 the `answerForScriptType.scriptType` is true even if no answer was provided.
> 
> This bug was introduced when we wanted to make the `preview` template option in Yo Office, as it didn't have a language and we didn't want to ask the language question. At that time, we didn't think about the `--ts` option and its impact on it.

So why does that work for the host prompt?  Also why is the conditional not needed for the project Name?  Also what about the js option?

We should be clear that each of the prompts is behaving correctly.  If that change broke one, it could have broken the others as well.

igor-ribeiiro:
I think I left the `host` prompt. I will be having a look into fixing it as well as other scenarios

igor-ribeiiro:
@millerds ready for the PR to be reviewed again

millerds:
Actually, I went back and looked more closely at that change that broke this and prompt settings and I think this if clauses that were added were unneeded (didn't notice it before).  They are being used to set a value when the question wasn't asked, but that should be handled by the "default" property of the prompt object.  That is why there is always an answer . . . if the question isn't asked the default value is used and we don't need to use the conditional to change it.  We should just remove the conditional.

igor-ribeiiro:
> lw

Yeah, the `if `wasn't necessary and I just removed it

millerds (approved):

PR #717, opened 2022-06-14
Adding shared template option

igor-ribeiiro:

---

    > 
    > * [X]  Yes
    > * [ ]  No

There will be a new option for the shared runtime template.

    > 
    > * [X]  Yes
    > * [ ]  No

The tutorial on how to create CFs need to be updated. It also needs to include shared vs legacy runtimes differences.

Tested creating a new yo office app and choosing the option shared template. The template is indeed the shared one.

4. **Platforms tested**:

    > * [X] Windows
    > * [ ] Mac

### DIFF
--- a/src/app/config/projectProperties.json
+++ b/src/app/config/projectProperties.json
@@ -129,6 +129,25 @@
                 "Excel"
             ]
         },
+        "excel-functions-shared": {
+            "displayname": "Excel Shared Runtime Custom Functions Add-in project",
+            "manifestPath": "manifest.xml",
+            "templates": {
+                "javascript": {
+                    "repository": "https://github.com/OfficeDev/Excel-Custom-Functions-JS",
+                    "branch": "shared-runtime-yo-office",
+                    "prerelease": "shared-runtime-yo-office-prerelease"
+                },
+                "typescript": {
+                    "repository": "https://github.com/OfficeDev/Excel-Custom-Functions",
+                    "branch": "shared-runtime-yo-office",
+                    "prerelease": "shared-runtime-yo-office-prerelease"
+                }
+            },
+            "supportedHosts": [
+                "Excel"
+            ]
+        },
         "teams-manifest": {
             "displayname": "Outlook Add-in with Teams Manifest (Developer preview)",
             "manifestPath": "manifest/manifest.json",
--- a/src/app/config/projectProperties.json
+++ b/src/app/config/projectProperties.json
@@ -23,17 +23,17 @@
                     "Project",
                     "Word"]
         },
-        "angular": {
-            "displayname": "Office Add-in Task Pane project using Angular framework",
+        "react": {
+            "displayname": "Office Add-in Task Pane project using React framework",
             "manifestPath": "manifest.xml",
             "templates": {
                 "javascript": {
-                    "repository": "https://github.com/OfficeDev/Office-Addin-TaskPane-Angular-JS",
+                    "repository": "https://github.com/OfficeDev/Office-Addin-TaskPane-React-JS",
                     "branch": "yo-office",
                     "prerelease": "yo-office-prerelease"
                 },
                 "typescript": {
-                    "repository": "https://github.com/OfficeDev/Office-Addin-TaskPane-Angular",
+                    "repository": "https://github.com/OfficeDev/Office-Addin-TaskPane-React",
                     "branch": "yo-office",
                     "prerelease": "yo-office-prerelease"
                 }
@@ -47,17 +47,17 @@
                 "Word"
             ]
         },
-        "react": {
-            "displayname": "Office Add-in Task Pane project using React framework",
+        "angular": {
+            "displayname": "Office Add-in Task Pane project using Angular framework",
             "manifestPath": "manifest.xml",
             "templates": {
                 "javascript": {
-                    "repository": "https://github.com/OfficeDev/Office-Addin-TaskPane-React-JS",
+                    "repository": "https://github.com/OfficeDev/Office-Addin-TaskPane-Angular-JS",
                     "branch": "yo-office",
                     "prerelease": "yo-office-prerelease"
                 },
                 "typescript": {
-                    "repository": "https://github.com/OfficeDev/Office-Addin-TaskPane-React",
+                    "repository": "https://github.com/OfficeDev/Office-Addin-TaskPane-Angular",
                     "branch": "yo-office",
                     "prerelease": "yo-office-prerelease"
                 }
@@ -93,55 +93,38 @@
                 "Word"
             ]
         },
-        "manifest": {
-            "displayname": "Office Add-in project containing the manifest only",
-            "manifestPath": "manifest.xml",
-            "templates": {
-                "manifestonly": {
-                    "repository": ""
-                }
-            },
-            "supportedHosts": [
-                "Excel",
-                "Onenote",
-                "Outlook",
-                "Powerpoint",
-                "Project",
-                "Word"
-            ]
-        },
-        "excel-functions": {
-            "displayname": "Excel Custom Functions Add-in project",
+        "excel-functions-shared": {
+            "displayname": "Excel Custom Functions using a Shared Runtime",
             "manifestPath": "manifest.xml",
             "templates": {
                 "javascript": {
                     "repository": "https://github.com/OfficeDev/Excel-Custom-Functions-JS",
-                    "branch": "yo-office",
-                    "prerelease": "yo-office-prerelease"
+                    "branch": "shared-runtime-yo-office",
+                    "prerelease": "shared-runtime-yo-office-prerelease"
                 },
                 "typescript": {
                     "repository": "https://github.com/OfficeDev/Excel-Custom-Functions",
-                    "branch": "yo-office",
-                    "prerelease": "yo-office-prerelease"
+                    "branch": "shared-runtime-yo-office",
+                    "prerelease": "shared-runtime-yo-office-prerelease"
                 }
             },
             "supportedHosts": [
                 "Excel"
             ]
         },
-        "excel-functions-shared": {
-            "displayname": "Excel Shared Runtime Custom Functions Add-in project",
+        "excel-functions": {
+            "displayname": "Excel Custom Functions using a JavaScript-only Runtime",
             "manifestPath": "manifest.xml",
             "templates": {
                 "javascript": {
                     "repository": "https://github.com/OfficeDev/Excel-Custom-Functions-JS",
-                    "branch": "shared-runtime-yo-office",
-                    "prerelease": "shared-runtime-yo-office-prerelease"
+                    "branch": "yo-office",
+                    "prerelease": "yo-office-prerelease"
                 },
                 "typescript": {
                     "repository": "https://github.com/OfficeDev/Excel-Custom-Functions",
-                    "branch": "shared-runtime-yo-office",
-                    "prerelease": "shared-runtime-yo-office-prerelease"
+                    "branch": "yo-office",
+                    "prerelease": "yo-office-prerelease"
                 }
             },
             "supportedHosts": [
@@ -160,6 +143,23 @@
             },
             "supportedHosts": [
                 "Outlook"
+            ]
+        },
+        "manifest": {
+            "displayname": "Office Add-in project containing the manifest only",
+            "manifestPath": "manifest.xml",
+            "templates": {
+                "manifestonly": {
+                    "repository": ""
+                }
+            },
+            "supportedHosts": [
+                "Excel",
+                "Onenote",
+                "Outlook",
+                "Powerpoint",
+                "Project",
+                "Word"
             ]
         }
     },
--- a/src/app/index.ts
+++ b/src/app/index.ts
@@ -287,28 +287,32 @@ module.exports = class extends yo {
 
   _configureProject(answerForProjectType, answerForScriptType, answerForHost, answerForName, isManifestProject, isExcelFunctionsProject): void {
     try {
+      const projType = _.toLower(this.options.projectType) || _.toLower(answerForProjectType.projectType)
+
       this.project = {
         folder: this.options.output || answerForName.name || this.options.name,
+        host: answerForHost.host
+          ? answerForHost.host
+          : this.options.host
+          ? this.options.host
+          : jsonData.getHostTemplateNames(projType)[0],
         name: this.options.name || answerForName.name,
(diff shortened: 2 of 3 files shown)


src/app/config/projectProperties.json
```diff
@@ -129,6 +129,25 @@
                 "Excel"
             ]
         },
+        "excel-functions-shared": {
+            "displayname": "Excel Shared Runtime Custom Functions Add-in project",
```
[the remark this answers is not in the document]
igor-ribeiiro: I don't think there's a name yet. I will keep this PR on hold until we get the name confirmation and then change when we have it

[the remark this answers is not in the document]
Rick-Kirkham: @alison-mk can help choose a name. If it were me, I'd like to avoid so many stacked nouns. I'd prefer breaking it up with a prepositional phrase. "Excel Custom Functions with a Shared Runtime project".

Rick-Kirkham (approved):

src/app/config/projectProperties.json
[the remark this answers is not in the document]
akrantz: "Excel Custom Functions using Shared Runtime"  ?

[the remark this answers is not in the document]
igor-ribeiiro: We can also change the normal Excel CFs name. Please let me know if we need to do that.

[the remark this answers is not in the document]
alison-mk: I like Ricky and Adam's suggestions. Could we also place Shared Runtime before JS-only in the list, to subtly indicate that it's the preferred method? @igor-ribeiiro Yes, I would also like to change the existing CF template name. I recommend: 

- Excel Custom Functions using a Shared Runtime
- Excel Custom Functions using a JavaScript-only Runtime

igor-ribeiiro:
Added the shared option before the manifest option and reordered the options for most used to least used.
The change https://github.com/OfficeDev/generator-office/pull/714/files caused a bug which the message wouldn't have the host in it:
image

So I also fixed that bug.

src/app/index.ts
```diff
@@ -196,6 +196,9 @@ module.exports = class extends yo {
         }
       ];
       const answerForScriptType = await this.prompt(askForScriptType);
+      if (!answerForScriptType.scriptType && !this.options.ts) {
```
millerds, line 199: We wanted to leave these off because the default should be set by the question settings.

src/app/index.ts
```diff
@@ -425,10 +426,12 @@ module.exports = class extends yo {
     this.log(`NOTE: ${chalk.bgGreen('Arguments')} must be specified in the order below, and ${chalk.bgMagenta('Options')} must follow ${chalk.bgGreen('Arguments')}.\n`);
     this.log(`  ${chalk.bgGreen('projectType')}:Specifies the type of project to create. Valid project types include:`);
     this.log(`    ${chalk.yellow('angular:')}  Creates an Office add-in using Angular framework.`);
-    this.log(`    ${chalk.yellow('excel-functions:')} Creates an Office add-in for Excel custom functions.  Must specify 'Excel' as host parameter.`);
+    this.log(`    ${chalk.yellow('excel-functions-shared:')} Creates an Office add-in for Excel custom functions using a Shared Runtime.  Must specify 'Excel' as host parameter.`);
```
millerds, line 429: Don't we automatically set the host parameter?  We wouldn't need this part of the message.

src/app/index.ts
```diff
@@ -425,10 +426,12 @@ module.exports = class extends yo {
...
+    this.log(`    ${chalk.yellow('excel-functions:')} Creates an Office add-in for Excel custom functions using a JavaScript-only Runtime.  Must specify 'Excel' as host parameter.`);
     this.log(`    ${chalk.yellow('jquery:')} Creates an Office add-in using Jquery framework.`);
     this.log(`    ${chalk.yellow('manifest:')} Creates an only the manifest file for an Office add-in.`);
     this.log(`    ${chalk.yellow('react:')} Creates an Office add-in using React framework.\n`);
+    this.log(`    ${chalk.yellow('teams-manifest:')} Creates Outlook Add-in with Teams Manifest (Developer preview).\n`);
```
millerds, line 434: Why is this listed as an add?  Shouldn't it already be there?

millerds (changes requested):

src/app/index.ts
igor-ribeiiro: I guess it's because the messages are a bit different between themselves. We would need to add a new parameter to all hosts (like `displayMessage`) if we would like to keep current messages.

src/app/index.ts
millerds: I'm not seeing that in the documentation and if it really didn't have the default answer then wouldn't there be an error when trying to create the project since there wouldn't be an answer (i.e. something like a null exception)?

millerds (changes requested):

igor-ribeiiro:
@millerds "I'm not seeing that in the documentation and if it really didn't have the default answer then wouldn't there be an error when trying to create the project since there wouldn't be an answer (i.e. something like a null exception)?"

It is null that gets returned (or something close to it). We don't get an error because we check if it exists:
```
        scriptType: answerForScriptType.scriptType ? answerForScriptType.scriptType : this.options.ts ? typescript : javascript
```

The `default` is only the value that will get chosen when you press "enter", without choosing a particular option.

millerds:
So if that line is taking care of the question not asked case, then there is no need for that if block.  It looks like the bug you are saying you are fixing is not with the language chosen, but rather the host (that is what is missing from the message).  The host also has a fall back in the same place as the language, but that isn't using the ?: case but rather the || syntax and probably behaves differently.

The point is that we shouldn't need two places to fall back onto and I prefer the _configureProject method with a simple value precedent check rather than having to maintain multiple conditionals with the same tests (the one for weather or not the question should be asked, and the same one for weather or not the question was asked.

igor-ribeiiro:
> Member

Sure, we can handle all these cases handling at the same place. I made some changes so that all cases are covered in a single line and hopefully it is not too confusing.

src/app/index.ts
```diff
@@ -295,12 +290,12 @@ module.exports = class extends yo {
...
         folder: this.options.output || answerForName.name || this.options.name,
         name: this.options.name || answerForName.name,
-        host: this.options.host || answerForHost.host,
         projectType: _.toLower(this.options.projectType) || _.toLower(answerForProjectType.projectType),
         isManifestOnly: isManifestProject,
         isExcelFunctionsProject: isExcelFunctionsProject,
-        scriptType: answerForScriptType.scriptType ? answerForScriptType.scriptType : this.options.ts ? typescript : javascript
       };
+      this.project.host = answerForHost.host ? answerForHost.host : this.options.host ? this.options.host : jsonData.getHostTemplateNames(this.project.projectType)[0];
+      this.project.scriptType = answerForScriptType.scriptType ? answerForScriptType.scriptType : this.options.ts ? typescript : this.options.js ? this.options.js : jsonData.getSupportedScriptTypes(this.project.projectType)[0];
```
millerds, line 298: I think the "this.options.js ? this.options.js" is intended to be "this.options.js ? javascript"

Could also use some formatting to encourage better readability rather than all on one line.

src/app/index.ts
```diff
@@ -29,6 +29,7 @@ let language;
 const manifest = 'manifest';
 const sso = 'single-sign-on';
 const typescript = `TypeScript`;
+let jsonData;
```
millerds, line 32: Why pull this out of the try block?

src/app/index.ts
```diff
@@ -295,12 +290,12 @@ module.exports = class extends yo {
       this.project = {
         folder: this.options.output || answerForName.name || this.options.name,
         name: this.options.name || answerForName.name,
-        host: this.options.host || answerForHost.host,
         projectType: _.toLower(this.options.projectType) || _.toLower(answerForProjectType.projectType),
         isManifestOnly: isManifestProject,
         isExcelFunctionsProject: isExcelFunctionsProject,
-        scriptType: answerForScriptType.scriptType ? answerForScriptType.scriptType : this.options.ts ? typescript : javascript
       };
+      this.project.host = answerForHost.host ? answerForHost.host : this.options.host ? this.options.host : jsonData.getHostTemplateNames(this.project.projectType)[0];
```
millerds, line 297: Why is this moved out the block it was in?

millerds (changes requested):

src/app/index.ts
igor-ribeiiro: It is being used on another function. So I wanted to let this other function `configureProject` also have access to `jsonData`

src/app/index.ts
```diff
@@ -289,12 +290,17 @@ module.exports = class extends yo {
       this.project = {
         folder: this.options.output || answerForName.name || this.options.name,
         name: this.options.name || answerForName.name,
-        host: this.options.host || answerForHost.host,
         projectType: _.toLower(this.options.projectType) || _.toLower(answerForProjectType.projectType),
         isManifestOnly: isManifestProject,
         isExcelFunctionsProject: isExcelFunctionsProject,
-        scriptType: answerForScriptType.scriptType ? answerForScriptType.scriptType : this.options.ts ? typescript : javascript
       };
+      this.project.host = answerForHost.host ? answerForHost.host
```
millerds, line 297: Why not leave this in the block using "host: answerforHost.host . . ." with everything else (you don't need to use "this.project.")?

src/app/index.ts
```diff
@@ -289,12 +290,17 @@ module.exports = class extends yo {
...
         name: this.options.name || answerForName.name,
-        host: this.options.host || answerForHost.host,
         projectType: _.toLower(this.options.projectType) || _.toLower(answerForProjectType.projectType),
         isManifestOnly: isManifestProject,
         isExcelFunctionsProject: isExcelFunctionsProject,
-        scriptType: answerForScriptType.scriptType ? answerForScriptType.scriptType : this.options.ts ? typescript : javascript
       };
+      this.project.host = answerForHost.host ? answerForHost.host 
+        : this.options.host ? this.options.host 
+        : jsonData.getHostTemplateNames(this.project.projectType)[0];
```
millerds, line 299: If this is coming from outside of this method then we should check that it's not empty.

millerds (changes requested):

src/app/index.ts
millerds: They set projectType inside the block as well.  It's a big deal but I would add 'const projType = _.toLower(this.options.projectType) || _.toLower(answerForProjectType.projectType)' before the block then create everything inside the block using the projType variable.

src/app/index.ts
```diff
@@ -287,20 +287,35 @@ module.exports = class extends yo {
...
   _configureProject(answerForProjectType, answerForScriptType, answerForHost, answerForName, isManifestProject, isExcelFunctionsProject): void {
     try {
+      const projType = _.toLower(this.options.projectType) || _.toLower(answerForProjectType.projectType)
+
       this.project = {
         folder: this.options.output || answerForName.name || this.options.name,
+        host: answerForHost.host
+          ? answerForHost.host
+          : this.options.host
+          ? this.options.host
```
millerds, line 297: Add a new indent level here.

src/app/index.ts
```diff
@@ -287,20 +287,35 @@ module.exports = class extends yo {
...
+          : jsonData.getHostTemplateNames(projType)[0],
         name: this.options.name || answerForName.name,
-        projectType: _.toLower(this.options.projectType) || _.toLower(answerForProjectType.projectType),
+        projectType: projType,
+        scriptType: answerForScriptType.scriptType
+          ? answerForScriptType.scriptType
+          : this.options.ts
+          ? typescript
```
millerds, line 304: Add an indent level here.

src/app/index.ts
```diff
@@ -287,20 +287,35 @@ module.exports = class extends yo {
...
+          : this.options.js
+          ? javascript
```
millerds, line 306: Add an indent level here.

millerds (changes requested):

src/app/index.ts
igor-ribeiiro: I'm using the prettier here. So those two spaces are the recommended amount.

src/app/index.ts
igor-ribeiiro: Same as other comment, I'm using `format selection` from vscode

src/app/index.ts
igor-ribeiiro: Using vscode format selection

src/app/index.ts
```diff
@@ -287,20 +287,35 @@ module.exports = class extends yo {
...
       };
-      this.project.host = answerForHost.host ? answerForHost.host 
-        : this.options.host ? this.options.host 
-        : jsonData.getHostTemplateNames(this.project.projectType)[0];
-      this.project.scriptType = answerForScriptType.scriptType ? answerForScriptType.scriptType
-        : this.options.ts ? typescript
-        : this.options.js ? javascript
-        : jsonData.getSupportedScriptTypes(this.project.projectType)[0];
+
+      if (!this.project.host) {
```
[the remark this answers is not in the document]
igor-ribeiiro: It should never be null unless there is an error with our code. I will just check for `jsonData` null

millerds (changes requested):

src/app/index.ts
```diff
@@ -295,7 +295,7 @@ module.exports = class extends yo {
           ? answerForHost.host
           : this.options.host
           ? this.options.host
-          : jsonData.getHostTemplateNames(projType)[0],
+          : jsonData?.getHostTemplateNames(projType)[0],
```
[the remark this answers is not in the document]
igor-ribeiiro: It would get host = `null` which would error out later on the code in the `npm run convertToSingleHost` line. 
JSON data is currently never null. If it were to be null, then there is an with our code itself not related to a client's path of choices.

In that case, do you think throwing an error if it is null would help? Ideally, this error never appears, but it might happen if we make mistakes in our code in the future.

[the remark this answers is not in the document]
millerds: It's probably OK as is.  You are right in that it's a safety for us rather than customers.  I just wanted to make sure there wouldn't be some other cryptic error or silent failure if null was encountered.

millerds (dismissed):

igor-ribeiiro:
@millerds could you review it again? Needed to comment some SSO tests that were added back during the merge

millerds (approved):

PR #794, opened 2023-11-28
Add one step samples for Word and Excel in Yeoman

richard6094:

---

    > 
    > * [ -]  Yes
    > * [ ]  No

There will be 2 more options for users' choices: 
image
Users could 1-step to get the sample add-in code, edit it and automatically launch.
2. **Do these changes impact *documentation*?** (e.g., a tutorial on https://learn.microsoft.com/en-us/office/dev/add-ins/overview/office-add-ins)
    > 
    > * [- ]  Yes
    > * [ ]  No

There will be an one-line command added for users to use the scripts in yeoman.

We have tested the parameters, options and error handlings
   

4. **Platforms tested**:

    > * [ - ] Windows
    > * [ ] Mac


src/app/config/Sample_Hello_world_script.ts
```diff
@@ -0,0 +1,187 @@
...
+                  // Step 3: Provide user the command to side-load add-in directly 
+                    console.log('Step [3/3]: Automatically launch add-in directly...');
+                    spinner.start();
+                    const crt_path = '.office-addin-dev-certs\\localhost.crt';
+                    const key_path = '.office-addin-dev-certs\\localhost.key';
+                    const full_path_crt = homeDirectory + '\\' + crt_path;
+                    const full_path_key = homeDirectory + '\\' + key_path;
+
+                    shell.cd('./Samples/hello-world/excel-hello-world');
+                    shell.exec('npm install -g office-addin-debugging', {async:true}, (code, stdout, stderr) => {
```
millerds, line 73: This should be part of the project and not installed seperately.

We also should not be installing things globally . . . we're targeted at a project and not modifying the general user environment.

src/app/config/Sample_Excel_Word.ts
```diff
@@ -0,0 +1,176 @@
...
+
+        console.log('Welcome to experience this Office add-in sample!');
+
+        // Step 1: Get sample code
+        console.log('Step [1/3]: Getting sample code...');
+        let spinner = new Spinner('Processing.. %s');
+        spinner.setSpinnerString('|/-\\');
+        spinner.start();
+
+        shell.exec('git clone https://github.com/OfficeDev/Excel-Scenario-based-Add-in-Samples.git', {async:true}, (code, stdout, stderr) => {
```
millerds, line 31: By cloning the project you are keeping a user connection with the original template/sample repo . . . which is not what we're trying to do here.  We're creating a new project that is not connected back to a template repo.  We also already have code in this plugin that accomplishes that . . . this is taking a completely divergent path.

src/app/index.ts
```diff
@@ -362,6 +451,13 @@ module.exports = class extends yo {
     });
   }
 
+  _postOpenSampleHints(): void {
```
millerds, line 454: We don't need to special case general information.

tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
     "compilerOptions": {
+        "allowJs": true,
```
millerds, line 3: Why is this needed?

millerds (changes requested):
This seems to diverge quite a bit from the design and intent of this plugin.  I samples case seems to do a lot of special casing to get unique behavior and experience and as a result we get mix of yeoman functionality and a custom cmd experience.  You've introdued the use of git, cloning instead of copying the source code, and extra user prompts . . . all of with are similar to but the the same as existing behavior.

You also have a lot of code that looks like it is mostly the same, but is different in only a few simple ways.  This should be made generic and parameterized instead of trying to maintain multiple copies of the same thing.

The basic idea of generating samples vs. blank projects needs to be thought of more thoroughly to come up with a long term design.

src/app/index.ts
richard6094: > We don't need to special case general information.

Can we just put the special case here temporarily for sample add-ins?

src/app/config/Sample_Excel_Word.ts
```diff
@@ -0,0 +1,176 @@
...
+                is_vscode_installed = true;
+                shell.exec('code -g README.md:22');
+            } else {
+                console.log('Visual Studio Code is not installed on your machine.');
+                shell.exec('start Mail-Merge-Sample-Add-in');
+            }
+
+            console.log('Step [2/3] completed!');
+            
+            // Ask user if sample Add-in automatic launch is needed
```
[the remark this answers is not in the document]
millerds: This is shipping code.  Evey if you want to experiment with something it would have to be in ship quality.  Note that we don't have the option to flight something and then just turn it off.

src/app/config/Sample_Excel_Word.ts
```diff
@@ -0,0 +1,176 @@
...
+            let auto_launch_answer = false;
+            rl.question('Do you want to continue with some operations? (Y/N)\n', (answer) => {
+                if (answer.trim().toLowerCase() == 'y') {
+                  // Continue with the operations
+                  // Step 3: Provide user the command to side-load add-in directly 
+                    console.log('Step [3/3]: Automatically side-load add-in directly...');
+                    spinner.start();
+
+                    shell.cd('./Mail-Merge-Sample-Add-in');
+                    shell.exec('npm install', {async:true}, (code, stdout, stderr) => {
```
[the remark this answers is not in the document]
millerds: This is shipping code.  Even if you want to experiment with something it would have to be in ship quality.  Note that we don't have the option to flight something and then just turn it off.

The generator doesn't distinguish between the list of options and what it generates . . . how are you not displaying it and yet using the generator.

src/app/index.ts
millerds: This is shipping code.  Even "temporary" things need to be ship quality.

PR #801, opened 2024-01-08
Upgrade axios and other packages to address vulnerabilities

kwame-ansong:
Upgrade packages that depend on axios to address vulnerabilties.


millerds:
I think the teams-manifest package go a fully fixed version published within the last 24 hours.  Can you run audit fix again to get the full fix?

PR #857, opened 2025-02-19
Update template name for JSON preview label

millerds:
    > 
    > * [x]  Yes
    > * [ ]  No

2. **Do these changes impact *documentation*?** (e.g., a tutorial on https://learn.microsoft.com/en-us/office/dev/add-ins/overview/office-add-ins)
    > 
    > * [ ]  Yes
    > * [x]  No

Ran automated tests and visually checked yo output.

4. **Platforms tested**:

    > * [x] Windows
    > * [ ] Mac

### DIFF
--- a/src/app/config/projectProperties.json
+++ b/src/app/config/projectProperties.json
@@ -47,7 +47,7 @@
             }
         },
         "taskpane-preview": {
-            "displayname": "Excel, PowerPoint, or Word Task Pane with unified manifest for Microsoft 365 (preview)",
+            "displayname": "Excel, PowerPoint, amd/or Word Task Pane with unified manifest for Microsoft 365 (preview)",
             "templates": {
                 "typescript": {
                     "repository": "https://github.com/OfficeDev/Office-Addin-TaskPane",
--- a/src/app/config/projectProperties.json
+++ b/src/app/config/projectProperties.json
@@ -47,7 +47,7 @@
             }
         },
         "taskpane-preview": {
-            "displayname": "Office Add-in Task Pane project (preview)",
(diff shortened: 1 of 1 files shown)


src/app/config/projectProperties.json
```diff
@@ -47,7 +47,7 @@
             }
         },
         "taskpane-preview": {
-            "displayname": "Office Add-in Task Pane project (preview)",
+            "displayname": "Excel, PowerPoint, or Word Task Pane with unified manifest for Microsoft 365 (preview)",
```
akrantz, line 50: Should this also include Outlook?    Can this use "and/or" since it can support more than one ?

millerds: This was the string provided by @Rick-Kirkham.  Technically it also supports Outlook, but that should be the same as the shipping taskpane template and we don't give you the option of just Outlook for this one.  It supports "all" which give us all 4 hosts together, but not a subset of all that is greater than a single host.  We can't put all of that into the name, so you have to go further to see your actual options.  This change was just to distinguish it from the base Taskpane option that isn't preview.

Once the support comes out of preview, it will fall back into the base Taskpane option.  It would take extra work here and a big template change to offer a combinatorial option.

Rick-Kirkham: I don't object to using "and/or".

Rick-Kirkham (dismissed):

src/app/config/projectProperties.json
Rick-Kirkham: "Excel, PowerPoint, amd/or Word Task Pane with unified manifest for Microsoft 365 (preview)"

Rick-Kirkham (approved):

